PR #71, opened 2021-01-13
Refactor PageInfo and RestClient

thecodepixi:

The previous `PageInfo` class and the `RestClient` class had a true dependency cycle. In reviewing with the team we realized the `PageInfo` class was unnecessary and its functionality could be refactored and replicated inside of the `RestClient` class itself. 

- moves `PageInfo` interface to `rest_client/types`
- adds `nextPage` and `prevPage` objects (as `GetRequestParams` type) to `PageInfo` interface
- adds `nextPage` and `prevPage` params objects to `ret.pageInfo` during `RestClient.request`
- removes the `PageInfo` class, tests, and types

Example of how this change works in practice (with impractical pseudo path/fields): 

```ts
const return = client.get({path: 'something', fields: ['some', 'fields'], limit: 10} 
const nextPage = client.get(return.pageInfo.nextPage)
```

## Type of change


FredyC:
So it seems this and #30 have completely escaped a documentation update? Or am I missing some mention of this?

PR #103, opened 2021-02-09
Enable probot/stale

mkevinosullivan:

Enables [probot-stale](https://github.com/probot/stale) on this repo.

Adds `.github/stale.yml`


andyw8:
I don't think this was active yet - I clicked through on https://github.com/apps/stale so we might see some activity now.

PR #115, opened 2021-02-25
Minor doc changes

### LINKED ISSUE #114: Add keywords to `package.json`
## Overview/summary

Add keywords to `package.json` to aid discoverability.

mkevinosullivan:

Some minor documentation/text updates needed based on post-release feedback

Changing the `description` attribute in `package.json`
Removing `TypeScript` as needed knowledge to use the library.
Adds keywords to `package.json` (fixes #114)

## Type of change

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@shopify/shopify-api",
   "version": "1.0.0",
-  "description": "Shopify TypeScript API to support core API functionality (auth, graphql proxy, webhooks)",
+  "description": "Shopify Admin API Library for Node - accelerate development with support for authentication, graphql proxy, webhooks",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "scripts": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -26,6 +26,13 @@
   "homepage": "https://github.com/Shopify/shopify-node-api",
   "author": "Shopify Inc.",
   "license": "MIT",
+  "keywords": [
+    "shopify",
+    "node",
+    "jwt",
+    "typescript",
+    "javascript"
```
paulomarg, line 34: I'm wondering about the keywords a bit - does it make sense to mention typescript / javascript / node?

I'd consider adding something like `shopify, app, graphql, rest, webhook` - does that make sense?

mkevinosullivan: I suppose it doesn't make sense for `typescript`, `javascript` or `node`, since it's hosted on `npmjs.org` anyway!  Maybe the `node` one to distinguish it from something that works in browsers?

The others make sense, will update.

paulomarg: Good point about `node`, let's keep that!

paulomarg (approved):

package.json
thecodepixi: Would "Shopify API Library" make sense in this list as well? Sort of how `themekit` has the keyword "themekit" as a direct reference to itself?

thecodepixi: Also, FWIW, the existing 3rd party libraries for node seem to mostly have just "Shopify" and "API" as their keywords 🤷 Maybe less is more?

PR #117, opened 2021-03-01
Add January 2021 API version

### LINKED ISSUE #116: Api Versions are out of date
# Issue summary

Using `import Shopify, {ApiVersion} from '@shopify/shopify-api';`
When I print out the available ApiVersions, I get an old list:
```
ApiVersions: {
  July19: '2019-07',
  October19: '2019-10',
  January20: '2020-01',
  April20: '2020-04',
  July20: '2020-07',
  October20: '2020-10',
  Unstable: 'unstable',
  Unversioned: 'unversioned'
}
```

Seems that these are cached values.  How do I get the latest API versions to load?

paulomarg:

Fixes #116 

The `2021-01` API version was missing from the version options.

Adding the missing version.

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 - Minor text/doc changes
+- Added `2021-01` API version to enum. [#117](https://github.com/shopify/shopify-node-api/pull/117)
+
 ## [1.0.0]
 
 - Initial public release
--- a/src/base_types.ts
+++ b/src/base_types.ts
@@ -22,6 +22,7 @@ export enum ApiVersion {
   April20 = '2020-04',
   July20 = '2020-07',
   October20 = '2020-10',
+  January21 = '2021-01',
   Unstable = 'unstable',
   Unversioned = 'unversioned',
 }


mllemango (approved):
did we want to implement something similar to the python library where we can automatically fetch the API versions? https://github.com/Shopify/shopify_python_api/blob/master/shopify/api_version.py#L30

paulomarg:
Fetching the versions is interesting, but I don't think we can define a Typescript type around a network request - the type needs to be defined at transpilation time, before we have a chance to run the request.

furnaceX:
> Fetching the versions is interesting, but I don't think we can define a Typescript type around a network request - the type needs to be defined at transpilation time, before we have a chance to run the request.

I had read that the Ruby shopify-api package so I had expected that this package worked similarly.  Perhaps there is a method that can be adopted from that package?
https://github.com/Shopify/shopify_api#breaking-change-notice-for-version-800

> Version 7.0.0 introduced ApiVersion, and known versions were hardcoded into the gem. Manually defining API versions is no longer required for versions not listed in the gem.

paulomarg:
I don't think it would work as far as defining a type goes, since I don't think we'd be able to create a type programatically (Typescript types are 'gone' by the time the code is running).

As someone mentioned on the original issue it should be possible to just use a plain string if the option isn't there. The type just translates that to a more readable format.

We can look into this in the future, as long as people are able to work around missing values in the enum in some way (in this case, using a plain string). If that fails for some reason we can address it as a bug, otherwise it'll be something for the future.

PR #118, opened 2021-03-01
[Feature] Export Shopify const as an alias to use with ES modules

stefanosala:

When using this lib in a ES module context (e.g. Node v15) you need to use the lib like this:

```javascript
import Shopify from '@shopify/shopify-api';

await Shopify.default.Clients.Rest(...);
```

With this change you can just import the `Shopify` variable directly:

```javascript
import { Shopify } from '@shopify/shopify-api';

await Shopify.Clients.Rest(...);
```

## Type of change

## Checklist

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -6,7 +6,7 @@ import ShopifyClients from './clients';
 import ShopifyUtils from './utils';
 import ShopifyWebhooks from './webhooks';
 
-const Shopify = {
+export const Shopify = {
   Context,
   Auth: ShopifyAuth,
   Session: ShopifySession,


stefanosala:
I'm honestly not sure if this is the right way of doing this, sorry! :)

paulomarg:
Hey @stefanosala, thank you for contributing! I think your change makes sense, I don't see any problems with it.

However, could you please clone the original repo, and create a PR from a branch in there? That will run our CI checks to make sure everything is in order before we merge your changes!

stefanosala:
@paulomarg I'm not sure I can push a branch on your original repo?

paulomarg:
Right @stefanosala, sorry about that! We've tweaked the checks so that they run on forks, could you please re-create this PR so we can run them?

stefanosala:
It should be all done in this pr :)

paulomarg (approved):

paulomarg:
Merged, thanks for bearing with us @stefanosala!

PR #119, opened 2021-03-01
Allow loading offline sessions from loadCurrentSession

paulomarg:

Currently, the `loadCurrentSession` method doesn't allow loading offline sessions from the request, which is desired for some apps.

Adding an optional parameter (defaulting to online) to allow returning offline sessions as well as online ones.

## Type of change

## Checklist

### DIFF
--- a/docs/usage/oauth.md
+++ b/docs/usage/oauth.md
@@ -112,16 +112,16 @@ You can use the `Shopify.Utils.loadCurrentSession()` method to load an online se
 
 As mentioned in the previous sections, you can use the OAuth methods to create both offline and online sessions. Once the process is completed, the session will be stored as per your `Context.SESSION_STORAGE` strategy, and can be retrieved with the below utitilies.
 
-- To load an online session:
+- To load a session, you can use the following method. You can load both online and offline sessions from the current request / response objects.
 ```ts
-await Shopify.Utils.loadCurrentSession(request, response)
+await Shopify.Utils.loadCurrentSession(request, response, isOnline);
 ```
-- To load an offline session:
+- If you need to load a session for a background job, you can get offline sessions directly from the shop.
 ```ts
-await Shopify.Utils.loadOfflineSession(shop)
+await Shopify.Utils.loadOfflineSession(shop);
 ```
 
-The library supports creating both offline and online sessions for the same shop, so it is up to the app to call the appropriate loading method depending on its needs.
+**Note**: the `loadOfflineSession` method does not perform any validations on the `shop` parameter. You should avoid calling it from user inputs or URLs.
 
 ## Detecting scope changes
 
--- a/src/auth/oauth/oauth.ts
+++ b/src/auth/oauth/oauth.ts
@@ -224,10 +224,15 @@ const ShopifyOAuth = {
   /**
    * Extracts the current session id from the request / response pair.
    *
-   * @param request HTTP request object
+   * @param request  HTTP request object
    * @param response HTTP response object
+   * @param isOnline Whether to load online (default) or offline sessions (optional)
    */
-  getCurrentSessionId(request: http.IncomingMessage, response: http.ServerResponse): string | undefined {
+  getCurrentSessionId(
+    request: http.IncomingMessage,
+    response: http.ServerResponse,
+    isOnline = true,
+  ): string | undefined {
     let currentSessionId: string | undefined;
 
     if (Context.IS_EMBEDDED_APP) {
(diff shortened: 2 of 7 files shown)


docs/usage/oauth.md
```diff
@@ -112,16 +112,16 @@ You can use the `Shopify.Utils.loadCurrentSession()` method to load an online se
...
 ```
-- To load an offline session:
+- If you need to load a session for a background job, you can get offline sessions directly from the shop.
 ```ts
-await Shopify.Utils.loadOfflineSession(shop)
+await Shopify.Utils.loadOfflineSession(shop);
 ```
 
-The library supports creating both offline and online sessions for the same shop, so it is up to the app to call the appropriate loading method depending on its needs.
+**Note**: the `loadOfflineSession` method does not perform any validations on the `shop` parameter. You should never called it from user inputs or URLs.
```
mkevinosullivan, line 124:
```suggestion
**Note**: the `loadOfflineSession` method does not perform any validations on the `shop` parameter. You should avoid calling it from user inputs or URLs.
```

src/auth/oauth/oauth.ts
```diff
@@ -224,10 +224,15 @@ const ShopifyOAuth = {
   /**
    * Extracts the current session id from the request / response pair.
    *
-   * @param request HTTP request object
+   * @param request  HTTP request object
    * @param response HTTP response object
+   * @param isOnline Whether to load online or offline sessions
```
mkevinosullivan, line 229:
```suggestion
   * @param isOnline Whether to load online (default) or offline sessions (optional)
```

src/utils/delete-current-session.ts
```diff
@@ -7,16 +7,18 @@ import * as ShopifyErrors from '../error';
 /**
  * Finds and deletes the current user's session, based on the given request and response
  *
- * @param req Current HTTP request
- * @param res Current HTTP response
+ * @param request  Current HTTP request
+ * @param response Current HTTP response
+ * @param isOnline Whether to load online or offline sessions
```
mkevinosullivan, line 12:
```suggestion
 * @param isOnline Whether to load online (default) or offline sessions (optional)
```

src/utils/load-current-session.ts
```diff
@@ -7,16 +7,18 @@ import {Session} from '../auth/session';
 /**
  * Loads the current user's session, based on the given request and response.
  *
- * @param req Current HTTP request
- * @param res Current HTTP response
+ * @param request  Current HTTP request
+ * @param response Current HTTP response
+ * @param isOnline Whether to load online or offline sessions
```
mkevinosullivan, line 12:
```suggestion
 * @param isOnline Whether to load online (default) or offline sessions (optional)
```

src/utils/test/load-current-session.test.ts
```diff
@@ -74,7 +75,7 @@ describe('loadCurrentSession', () => {
     const session = new Session(`test-shop.myshopify.io_${jwtPayload.sub}`);
     await expect(Context.SESSION_STORAGE.storeSession(session)).resolves.toEqual(true);
 
-    await expect(loadCurrentSession(req, res)).resolves.toEqual(session);
+    await expect(loadCurrentSession(req, res, true)).resolves.toEqual(session);
```
mkevinosullivan, line 78: Is the `true` necessary, since it's the default?

src/auth/oauth/oauth.ts
```diff
@@ -239,13 +244,19 @@ const ShopifyOAuth = {
...
+        if (isOnline) {
+          currentSessionId = this.getJwtSessionId(shop, jwtPayload.sub);
+        } else {
+          currentSessionId = this.getOfflineSessionId(shop);
+        }
       }
     }
 
     // We fall back to the cookie session to allow apps to load their skeleton page after OAuth, so they can set up App
     // Bridge and get a new JWT.
```
mkevinosullivan, line 257: This comment isn't quite correct - it's not just the post-OAuth, pre-JWT embedded scenario, it's also the non-embedded scenario, right?

src/utils/test/load-current-session.test.ts
paulomarg: Not particularly!

src/auth/oauth/oauth.ts
paulomarg: True - the comment applies to embedded apps when it falls back, but it always applies for non-embedded ones. Will reword it (hopefully we can remove it soon).

mkevinosullivan (approved):

PR #122, opened 2021-03-04
Run CI checks on both PRs and pushes

paulomarg:

We are currently not running CI checks on PRs from forks, which makes it harder to spot errors before we merge them.

Running the CI workflow on both PRs and pushes so we can cover external contributions as well.

## Checklist

- [ ] ~I have added a changelog entry, prefixed by the type of change noted above~
- [ ] ~I have added/updated tests for this change~

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -1,4 +1,4 @@
-on: push
+on: [ push, pull_request ]
 name: CI
 jobs:
   CI:


mllemango (approved):

.github/workflows/workflow.yml
```diff
@@ -1,4 +1,4 @@
-on: push
+on: [ push, pull_request ]
```
Paulinakhew, line 1:
```suggestion
on:
  push:
    branches-ignore:
      - master
  pull_request:
    branches:
      - master
```

Paulinakhew (changes requested):
Can you try out this way to avoid duplication?

.github/workflows/workflow.yml
Paulinakhew: I put a better solution on Slack

PR #123, opened 2021-03-04
Add params tables for GraphQL and Rest Client docs.

### LINKED ISSUE #121: Documentation improvements
## Overview/summary

Now that the library has been released, we noticed a couple of pain points that we can help remedy via documentation. This PR aggregates them so we can address them at once.

1. Document the methods and parameters for the REST and GraphQL clients (using the same table format from the `withSession` docs).
1. Add a concrete production-valid example of how to use `CustomSessionStorage` in the session notes. We could show how to implement the callbacks using Redis.
2.1. We may want to consider adding a few tips (and what-not-to-dos!) on using this feature to help developers choose their preferred method.

## Checklist

- [X] I have described this enhancement in a way that is actionable (if possible)

thecodepixi:

~Partially fixes 121~ 

Strikethru the above to keep docs issue open, but this does partially resolve some of the needed docs enhancement outlined in that issue. 

Adds the necessary tables to explain the available params and arguments for the `GraphQL` and `Rest` clients, as well as general improvements to the usage examples. 

## Type of change

~- [ ] Minor: New feature (non-breaking change which adds functionality)~
~- [ ] Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)~

## Checklist

~- [ ] I have added/updated tests for this change~
~- [ ] I have documented new APIs/updated the documentation for modified APIs (for public APIs)~

### DIFF
--- a/docs/usage/graphql.md
+++ b/docs/usage/graphql.md
@@ -2,11 +2,21 @@
 
 Once OAuth is complete, we can use the library's `GraphqlClient` to make requests to the GraphQL Admin API in a similar way. To do that, create an instance of `GraphqlClient` using the current shop URL and session `accessToken` in your app's endpoint.
 
+The `GraphQLClient`'s main method is `query`, which accepts a `GraphQLParams` object as its argument. `GraphQLParams` only requires the `data` parameter, but also optionally accepts `query`, `extraHeaders`, and `tries`:
+
+| Parameter      | Type                                | Notes                                                                                 | Required? | Default Value |
+| -------------- | ----------------------------------- | ------------------------------------------------------------------------------------- | :-------: | :-----------: |
+| `data`         | `Record<string, unknown> \| string` | Takes in either the query as a string, and an object containing a query and variables |   True    |     none      |
+| `query`        | `Record<string, string \| number>`  | An optional query object to be appended to the request                                |   False   |     none      |
+| `extraHeaders` | `Record<string, string \| number>`  | Any additional headers you want to send with your request                             |   False   |     none      |
+| `tries`        | `number`                            | The maximum number of times to retry the request _(must be >= 0)_                     |   False   |      `1`      |
+
 ```ts
-  const session = await Shopify.Utils.loadCurrentSession(req, res); // load the current session to get the `accessToken`
-  const client = new Shopify.Clients.Graphql(session.shop, session.accessToken); // GraphQLClient accepts the same arguments as RestClient
-  const products = await client.query({ // use client.query and pass your query as `data`
-    data: `{
+const session = await Shopify.Utils.loadCurrentSession(req, res); // load the current session to get the `accessToken`
+const client = new Shopify.Clients.Graphql(session.shop, session.accessToken); // GraphQLClient accepts the same arguments as RestClient
+const products = await client.query({
+  // use client.query and pass your query as `data`
+  data: `{
       products (first: 10) {
         edges {
           node {
@@ -16,9 +26,9 @@ Once OAuth is complete, we can use the library's `GraphqlClient` to make request
           }
         }
       }
-    }`
-  });
-  // do something with the returned data
+    }`,
+});
+// do something with the returned data
 ```
 
 [Back to guide index](../index.md)
--- a/docs/usage/graphql.md
+++ b/docs/usage/graphql.md
@@ -6,15 +6,15 @@ The `GraphQLClient`'s main method is `query`, which accepts a `GraphQLParams` ob
 
 | Parameter      | Type                                |Required? | Default Value | Notes                                                                                 |
 | -------------- | ----------------------------------- |:-------: | :-----------: | ------------------------------------------------------------------------------------- |
-| `data`         | `Record<string, unknown> \| string` |    True    |     none     | Takes in either the query as a string, and an object containing a query and variables |
-| `query`        | `Record<string, string \| number>`  |   False   |     none      | An optional query object to be appended to the request                                |
-| `extraHeaders` | `Record<string, string \| number>`  |   False   |     none      | Any additional headers you want to send with your request                             |
-| `tries`        | `number`                            |   False   |      `1`      | The maximum number of times to retry the request _(must be >= 0)_                     |
+| `data`         | `Record<string, unknown> \| string` |    True  |     none      | Takes in either the query as a string, and an object containing a query and variables |
+| `query`        | `Record<string, string \| number>`  |   False  |     none      | An optional query object to be appended to the request                                |
+| `extraHeaders` | `Record<string, string \| number>`  |   False  |     none      | Any additional headers you want to send with your request                             |
+| `tries`        | `number`                            |   False  |      `1`      | The maximum number of times to retry the request _(must be >= 0)_                     |
 
 ```ts
 // Load the current session to get the `accessToken`
 const session = await Shopify.Utils.loadCurrentSession(req, res);
-// GraphQLClient takes in the a the shop and the accessToken for that shop.
+// GraphQLClient takes in the shop url and the accessToken for that shop.
 const client = new Shopify.Clients.Graphql(session.shop, session.accessToken);
 // Use client.query and pass your query as `data`
 const products = await client.query({
--- a/docs/usage/graphql.md
+++ b/docs/usage/graphql.md
@@ -4,18 +4,20 @@ Once OAuth is complete, we can use the library's `GraphqlClient` to make request
 
 The `GraphQLClient`'s main method is `query`, which accepts a `GraphQLParams` object as its argument. `GraphQLParams` only requires the `data` parameter, but also optionally accepts `query`, `extraHeaders`, and `tries`:
 
-| Parameter      | Type                                | Notes                                                                                 | Required? | Default Value |
-| -------------- | ----------------------------------- | ------------------------------------------------------------------------------------- | :-------: | :-----------: |
-| `data`         | `Record<string, unknown> \| string` | Takes in either the query as a string, and an object containing a query and variables |   True    |     none      |
-| `query`        | `Record<string, string \| number>`  | An optional query object to be appended to the request                                |   False   |     none      |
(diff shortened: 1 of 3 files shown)


docs/usage/rest.md
```diff
@@ -2,16 +2,70 @@
...
+// do something with the returned data
+```
+
+**Perform a `post` request:**
+
+```ts
+const session = await Shopify.Utils.loadCurrentSession(req, res); // load the current session to get the `accessToken`
+const client = new Shopify.Clients.Rest(session.shop, session.accessToken); // create a new client for the specified shop
+const body = {
+  // post request body for a new product
```
thecodepixi, line 63: Should we provide an example of a possible request body, or is this not necessary? The docs are more about the usage of the specific client and not generally about "how to structure a post request", so I'm unsure.

docs/usage/graphql.md
```diff
@@ -2,11 +2,21 @@
 
 Once OAuth is complete, we can use the library's `GraphqlClient` to make requests to the GraphQL Admin API in a similar way. To do that, create an instance of `GraphqlClient` using the current shop URL and session `accessToken` in your app's endpoint.
 
+The `GraphQLClient`'s main method is `query`, which accepts a `GraphQLParams` object as its argument. `GraphQLParams` only requires the `data` parameter, but also optionally accepts `query`, `extraHeaders`, and `tries`:
+
+| Parameter      | Type                                | Notes                                                                                 | Required? | Default Value |
```
gfscott, line 7: I think a good rule is to always put the "Notes" column last, since it's often the largest chunk and in some contexts might push more critical information (like "Required?") out of the user's view

docs/usage/graphql.md
```diff
@@ -2,11 +2,21 @@
...
+| `query`        | `Record<string, string \| number>`  | An optional query object to be appended to the request                                |   False   |     none      |
+| `extraHeaders` | `Record<string, string \| number>`  | Any additional headers you want to send with your request                             |   False   |     none      |
+| `tries`        | `number`                            | The maximum number of times to retry the request _(must be >= 0)_                     |   False   |      `1`      |
+
 ```ts
-  const session = await Shopify.Utils.loadCurrentSession(req, res); // load the current session to get the `accessToken`
-  const client = new Shopify.Clients.Graphql(session.shop, session.accessToken); // GraphQLClient accepts the same arguments as RestClient
-  const products = await client.query({ // use client.query and pass your query as `data`
-    data: `{
+const session = await Shopify.Utils.loadCurrentSession(req, res); // load the current session to get the `accessToken`
```
gfscott, line 15: I don't think we have a set rule about this (and perhaps we should define one!) but I'd like to consider the ergonomics of using trailing comments in example code, as opposed to putting the comment on the line above, especially with longer lines:

```js
// comment
code
```
vs:
```js
code // comment
```

Dunno if anyone else has preferences/thoughts about this.

docs/usage/graphql.md
```diff
@@ -2,11 +2,21 @@
...
+const client = new Shopify.Clients.Graphql(session.shop, session.accessToken); // GraphQLClient accepts the same arguments as RestClient
+const products = await client.query({
+  // use client.query and pass your query as `data`
+  data: `{
```
gfscott, line 19: Similarly, we probably need a better standard for this: I would have put this comment above the `const products` line, to avoid having to interrupt the actual query declaration with a comment line.

```suggestion
// use client.query and pass your query as `data`
const products = await client.query({
  data: `{
```

docs/usage/rest.md
```diff
@@ -2,16 +2,70 @@
...
+## Methods
+
+The `RestClient` offers the 4 core request methods: `get`, `delete`, `post`, and `put`, . These methods each take in an associated `<method>RequestParams` object as their argument, each with a set of required and optional parameters. The tables below outline these params objects.
+
+**Get and Delete:**
+
+- The `get` method accepts a `GetRequestParams` object, which accepts the following parameters: `path`, `type`, `data`, `query`, `extraHeaders`, and `tries`.
+
+- **`GetRequestParams`:**
+  | Parameter | Type | Notes | Required? | Default Value |
```
gfscott, line 14: Same deal here, "Notes" as the last column, I think

docs/usage/rest.md
gfscott: Maybe not a full example, but just a comment and [link to the reference](https://shopify.dev/docs/admin-api/rest/reference/products/product#create-2021-01)?

gfscott (approved):
👍🏻 Looking good! Couple little notes and considerations

docs/usage/graphql.md
paulomarg: Personally, I prefer to put comments before the code, it can be harder to see which line the comment applies to, but it makes lines shorter and keep things visually aligned.

docs/usage/rest.md
```diff
@@ -2,16 +2,70 @@
...
+
+```ts
+const session = await Shopify.Utils.loadCurrentSession(req, res); // load the current session to get the `accessToken`
+const client = new Shopify.Clients.Rest(session.shop, session.accessToken); // create a new client for the specified shop
+const body = {
+  // post request body for a new product
+};
+await client.post({
+  path: 'products',
+  data: body,
```
paulomarg, line 67: We currently require the `type` param here.

paulomarg (commented):
This is great!!!

docs/usage/graphql.md
thecodepixi: Oh that's a really good point, thank you! I'll shift that over.

docs/usage/graphql.md
thecodepixi: I think there's a lot of debate about this in "actual" code as well. On the one hand, I see the inline/trailing comment as being more clearly tied to that line, or one of its arguments/behavior etc, but the above-line comments might be better for readability and line-length. My feeling on inline comments tends to depend on their overall length, I think. But having a set rule/convention for this would be great!

docs/usage/rest.md
```diff
@@ -2,16 +2,70 @@
...
+- **`GetRequestParams`:**
+  | Parameter | Type | Notes | Required? | Default Value |
+  | -------------- | ----------------------------------- | ------------------------------------------------------------------------------------- | :---------: | :-------------: |
+  | `path` | `string` |The requested API endpoint path | True | none |
+  | `data` | `Record<string, unknown> \| string` | The body of the request | False | none |
+  | `type` | `DataType` | The type of data being sent in the body of the request (`JSON`, `GraphQL`, `URLEncoded`) | False | none |
+  | `query` | `Record<string, string \| number>` | An optional query object to be appended to the request | False | none |
+  | `extraHeaders` | `Record<string, string \| number>` | Any additional headers you want to send with your request | False | none |
+  | `tries` | `number` | The maximum number of times to retry the request _(must be >= 0)_ | False | `1` |
+- The `delete` method accepts a `DeleteRequestParams` object, **which has an identical strucuture to `GetRequestParams`**.
```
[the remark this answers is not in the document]
thecodepixi: Good idea!

docs/usage/rest.md
thecodepixi: Thank you! 🤦

docs/usage/graphql.md
thecodepixi: @paulomarg Yeah I feel the same way. Is it obvious enough to just end the param name in a `?` to mark it as optional? So few of the params are required it feels like a waste of table space. I actually feel the same way about the `Default Value` column, but want to communicate the default for `tries` in an obvious way.

docs/usage/graphql.md
thecodepixi: @paulomarg I'm fine with before the code and proper sentence casing as well. Probably better from a presentational standpoint.

docs/usage/rest.md
```diff
@@ -2,16 +2,78 @@
...
+  | `type` | `DataType` | True | none | The type of data being sent in the body of the request (`JSON`, `GraphQL`, `URLEncoded`) |
+  | `query` | `Record<string, string \| number>` | False | none | An optional query object to be appended to the request |
+  | `extraHeaders` | `Record<string, string \| number>` | False | none | Any additional headers you want to send with your request |
+  | `tries` | `number` | False | `1` | The maximum number of times to retry the request _(must be >= 0)_ |
+
+## Usage Examples:
+
+We can run the code below in any endpoint where we have access to the request and response objects.
+
+**Perform a `get` request:**
```
paulomarg, line 43: Nit: use GET / POST rather than get / post?

docs/usage/rest.md
```diff
@@ -2,16 +2,78 @@
...
+  | `query` | `Record<string, string \| number>` | False | none | An optional query object to be appended to the request |
+  | `extraHeaders` | `Record<string, string \| number>` | False | none | Any additional headers you want to send with your request |
+  | `tries` | `number` | False | `1` | The maximum number of times to retry the request _(must be >= 0)_ |
+
+**Post and Put:**
+
+- The `post` method accepts a `PostRequestParams` object, which differs from `get` in that both the `type` and `data` parameters are required, in addition to the `path`.
+- The `put` method accepts a `PutRequestParams` object, **which has an identical structure to `PostRequestParams`**.
+-
+- **`PostRequestParams` / `PutRequestParams`:**
```
paulomarg, line 29: I like this much better, we should do it on the GraphQL file too! Couple of nits:
- We have an empty `-` line here, we should remove that
- For the titles of all of these subsections, should we use increasing header levels? (`###`, `####`, etc.)

docs/usage/rest.md
thecodepixi: Fair!

docs/usage/rest.md
thecodepixi: Thanks for catching the blank line! Not sure how I missed that. And +1 to the headers, will update those.

docs/usage/rest.md
thecodepixi: I think I went for the lowercase because it's the literal method (`client.get`) but using the uppercase is probably more clear and more conventional.

paulomarg (approved):
LGTM

PR #126, opened 2021-03-05
Allow plain objects in Custom `loadSession` method

thecodepixi:

Previously, we always expected the return from the `loadCallback` to already be an `instanceof Session`, but this can cause some issues in developing the `loadCallback` method for a custom storage solution, due to having to do some extra object manipulation. 

`CustomSessionStorage.loadSession` will now allow the `loadCallback` to return _either_ an `instanceof Session` or a plain object. If a plain object is returned, it will be used to instantiate a `new Session` which will then be returned. 

## Type of change

~- [ ] Minor: New feature (non-breaking change which adds functionality)~
~- [ ] Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)~

## Checklist

_NOTE: I have updated the type on the current `CustomSessionStorage` docs under `issues.md`, but will be adding to and updating this docs for `CustomSessionStorage` more thoroughly in a separate PR._

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,14 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Added
+
+### Fixed
+
+- Allow plain objects to be returned from the `loadCallback` on `CustomSessionStorage` [#126](https://github.com/shopify/shopify-node-api/pull/126)
+
 ## [1.1.0] - 2021-03-02
+
 - Minor text/doc changes
 - Added `2021-01` API version to enum. [#117](https://github.com/shopify/shopify-node-api/pull/117)
 - Allow retrieving offline sessions using `loadCurrentSession`. [#119](https://github.com/shopify/shopify-node-api/pull/119)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 ## Unreleased
 
 ### Added
+- Allow plain objects to be returned from the `loadCallback` on `CustomSessionStorage` [#126](https://github.com/shopify/shopify-node-api/pull/126)
 
 ### Fixed
 
-- Allow plain objects to be returned from the `loadCallback` on `CustomSessionStorage` [#126](https://github.com/shopify/shopify-node-api/pull/126)
 
 ## [1.1.0] - 2021-03-02
 
--- a/src/auth/session/storage/custom.ts
+++ b/src/auth/session/storage/custom.ts
@@ -37,7 +37,7 @@ export class CustomSessionStorage implements SessionStorage {
         return result;
       } else if (result instanceof Object && 'id' in result) {
         let session = new Session(result.id as string);
-        session = Object.assign(session, result);
+        session = {...session, ...result};
         return session;
       } else {
         throw new ShopifyErrors.SessionStorageError(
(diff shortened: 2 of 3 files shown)


src/auth/session/storage/custom.ts
```diff
@@ -35,6 +35,10 @@ export class CustomSessionStorage implements SessionStorage {
     if (result) {
       if (result instanceof Session) {
         return result;
+      } else if (result instanceof Object && 'id' in result) {
```
paulomarg, line 38: I like this but I wonder if it wouldn't be simpler to simply take in the JSON string? We could allow both but that seems a bit much.

What do you think?

thecodepixi: I'm not sure I have strong feeling for or against that, it would just require more checking. One thing it does raise in my mind is that we're maybe reaching into their storage implementation by doing that? The custom storage solution should handle its own data in the `loadCallback` and then give it to us to essentially "pass through" `loadSession`. By offering to manipulate stringified objects (or potentially non-parseable strings) we're combining their solution with our solution, in a way. But if parsing doesn't work, that's their own error, not ours.

paulomarg: That's fair - at some point we have to translate the external object to ours, but it does make it more flexible if we don't assume it will be JSON.

Even though I suspect JSON will be used in the vast majority of the cases, there's no reason not to support other methods as well.

thecodepixi: > there's no reason not to support other methods as well.

I can't tell if we're on the same page here? Do you mean we should add support for further data types in this method? My philosophy is "please make your data into this shape first" and have them pass in at least a JSON object. I think the manipulation of their data should be more delegated to their custom callback than to the library, personally. But since the library is what stipulates "make this a `Session`" it makes sense that we handle that step for them.

paulomarg: No, I was agreeing with you - as long as they give us an object that maps to what our session looks like, apps can do whatever they want with their objects. By 'other formats' I meant in how they would possibly serialize and de-serialize the object to store it in their database (like say a MongoDB object that's NOT converted to JSON, or something).

thecodepixi: Ah ok yes! Yeah, I think this way we're being as flexible as possible, because we don't really care what they do to store/access the data, as long as it comes back to us in the right shape.

CHANGELOG.md
```diff
@@ -7,7 +7,14 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Added
+
+### Fixed
+
+- Allow plain objects to be returned from the `loadCallback` on `CustomSessionStorage` [#126](https://github.com/shopify/shopify-node-api/pull/126)
```
paulomarg, line 14: Total nit: I think this is more of an addition than a fix, nothing was broken before.

src/auth/session/storage/custom.ts
```diff
@@ -35,6 +35,10 @@ export class CustomSessionStorage implements SessionStorage {
...
+        let session = new Session(result.id as string);
+        session = Object.assign(session, result);
```
paulomarg, line 40: I think the more 'modern' way of doing this would be (untested and unsure if the result will actually be a `Session`):
```suggestion
        session = {...session, ...result};
```

paulomarg (approved):
Minor nits but I'm good with it being merged as is.

CHANGELOG.md
thecodepixi: Fair point

src/auth/session/storage/custom.ts
thecodepixi: I'll try it out, but I'm not 100% sure this will work.

thecodepixi: TIL this totally does work this way.

kato-takaomi-ams: `session = {...session, ...result};`

I think the more type-strict way of doing this.
`session = Session.cloneSession(result, result.id);`

Issue #125, opened 2021-03-05
Passing custom ARN argument when registering EventBridge webhooks


rdunk:
Maybe more of a question as I'm assuming I may have misinterpreted the docs.

When registering EventBridge webhooks, the [`arn`](https://github.com/Shopify/shopify-node-api/blob/da4cfc200350bc1b90d6e271f1bdb875f7b5bb01/src/webhooks/registry.ts#L116) argument [looks like it's built](https://github.com/Shopify/shopify-node-api/blob/da4cfc200350bc1b90d6e271f1bdb875f7b5bb01/src/webhooks/registry.ts#L147) as a URL in the following format:

`https://${Context.HOST_NAME}${path}`

[This tutorial](https://shopify.dev/tutorials/manage-webhook-events-with-eventbridge?itcat=partner_blog&itterm=shopify_webhooks#4-create-webhook-subscriptions-for-eventbridge) suggests using the Partner event source ARN which as far as I can tell, for Shopify events, follow the format:

`arn:aws:events:region::event-source/aws.partner/shopify.com/id/name`

Is there a way to set the ARN argument manually, or am I missing something here? Thanks.

paulomarg:
This is a great point @rdunk, right now we're always using an HTTP address. We'll look into supporting other alternatives!

mkevinosullivan:
Closing this, as the PR supporting this is now merged.

Thanks @rdunk for the great contribution 👍🏻

PR #127, opened 2021-03-06
[WIP] [Feature] Support ARNs for EventBridge webhooks

rdunk:

This is a potential solution for the issue described in #125.

My understanding is that EventBridge webhooks must be registered with an ARN rather than a URL, so this introduces changes to facilitate passing an ARN as a path, or defining one via the `Context` object.

When constructing an address in the webhook register method, some conditional logic has been added to differentiate EventBridge vs HTTP subscriptions.

If `DeliveryMethod.EventBridge` is used, the address is assigned the verbatim value of the `path` option, rather than prefixing it with a protocol extension + `HOST_NAME`. In addition, `path` becomes optional for the EventBridge specific options object (see the updated `RegisterOptions` type), and if not explicitly provided the address will fallback to the additional `EVENTBRIDGE_ARN` property that has been added to the `ContextParams` type. 

To support ARNs, the deprecated top-level `callbackUrl` field on the GraphQL `WebhookSubscription` object has been removed and replaced with the new `endpoint` field. I believe this drops webhook support for API versions <=2020-04, which I understand may not be ideal.

## Type of change

## Checklist

### DIFF
--- a/src/base_types.ts
+++ b/src/base_types.ts
@@ -12,6 +12,7 @@ export interface ContextParams {
   SESSION_STORAGE?: SessionStorage;
   LOG_FILE?: string;
   USER_AGENT_PREFIX?: string;
+  EVENTBRIDGE_ARN: string;
 }
 
 export enum ApiVersion {
--- a/src/base_types.ts
+++ b/src/base_types.ts
@@ -12,7 +12,6 @@ export interface ContextParams {
   SESSION_STORAGE?: SessionStorage;
   LOG_FILE?: string;
   USER_AGENT_PREFIX?: string;
-  EVENTBRIDGE_ARN: string;
 }
 
 export enum ApiVersion {
--- a/src/utils/version-compatible.ts
+++ b/src/utils/version-compatible.ts
@@ -12,6 +12,5 @@ export default function versionCompatible(referenceVersion: ApiVersion, currentV
   const numericVersion = (version: string) => parseInt(version.replace('-', ''), 10);
   const current = numericVersion(currentVersion);
   const reference = numericVersion(referenceVersion);
-  console.log(current, reference, current >= reference);
   return current >= reference;
 }
--- a/src/utils/version-compatible.ts
+++ b/src/utils/version-compatible.ts
@@ -1,0 +1,17 @@
+import {Context} from '../context';
+import {ApiVersion} from '../base_types';
+
+/**
+ * Check if the current or optionally supplied version is compatible with a given version
+ */
+export default function versionCompatible(referenceVersion: ApiVersion, currentVersion: ApiVersion = Context.API_VERSION): boolean {
+  // Return true if not using a dated version
+  if (currentVersion === ApiVersion.Unstable || currentVersion === ApiVersion.Unversioned) {
+    return true;
+  }
+  const numericVersion = (version: string) => parseInt(version.replace('-', ''), 10);
+  const current = numericVersion(currentVersion);
+  const reference = numericVersion(referenceVersion);
+  console.log(current, reference, current >= reference);
+  return current >= reference;
+}
--- a/src/webhooks/registry.ts
+++ b/src/webhooks/registry.ts
@@ -292,4 +292,4 @@ const WebhooksRegistry: RegistryInterface = {
   },
 };
 
-export {WebhooksRegistry, RegistryInterface};
+export {WebhooksRegistry, RegistryInterface, buildCheckQuery, buildQuery};
--- a/src/webhooks/registry.ts
+++ b/src/webhooks/registry.ts
@@ -84,6 +84,12 @@ function versionSupportsEndpointField() {
   return ShopifyUtilities.versionCompatible(ApiVersion.July20);
 }
 
+function validateDeliveryMethod(deliveryMethod: DeliveryMethod) {
+  if (deliveryMethod === DeliveryMethod.EventBridge && !versionSupportsEndpointField()) {
+    throw new ShopifyErrors.UnsupportedClientType(`EventBridge webhooks are not supported in API version "${Context.API_VERSION}".`);
+  }
+}
+
 function buildCheckQuery(topic: string): string {
   const query = `{
     webhookSubscriptions(first: 1, topics: ${topic}) {
@@ -124,10 +130,7 @@ function buildQuery(
   deliveryMethod: DeliveryMethod = DeliveryMethod.Http,
   webhookId?: string,
 ): string {
-  if (deliveryMethod === DeliveryMethod.EventBridge && !versionSupportsEndpointField()) {
-    // No endpoint fallback exists for EventBridge, so throw early
-    throw new ShopifyErrors.UnsupportedClientType(`EventBridge webhooks are not supported in API version "${Context.API_VERSION}".`);
-  }
+  validateDeliveryMethod(deliveryMethod);
   let identifier: string;
   if (webhookId) {
     identifier = `id: "${webhookId}"`;
@@ -174,6 +177,7 @@ const WebhooksRegistry: RegistryInterface = {
     deliveryMethod = DeliveryMethod.Http,
     webhookHandler,
   }: RegisterOptions): Promise<RegisterReturn> {
+    validateDeliveryMethod(deliveryMethod);
     const client = new GraphqlClient(shop, accessToken);
     const address = deliveryMethod === DeliveryMethod.EventBridge
       ? path
--- a/src/webhooks/registry.ts
+++ b/src/webhooks/registry.ts
@@ -152,10 +152,9 @@ const WebhooksRegistry: RegistryInterface = {
     webhookHandler,
   }: RegisterOptions): Promise<RegisterReturn> {
     const client = new GraphqlClient(shop, accessToken);
-    const address =
-      deliveryMethod && deliveryMethod === DeliveryMethod.EventBridge
-        ? path
-        : `https://${Context.HOST_NAME}${path}`;
+    const address = deliveryMethod === DeliveryMethod.EventBridge
+      ? path
+      : `https://${Context.HOST_NAME}${path}`;
     const checkResult = await client.query({
       data: buildCheckQuery(topic),
     });
--- a/src/webhooks/registry.ts
+++ b/src/webhooks/registry.ts
@@ -138,7 +138,7 @@ function buildQuery(
   let mutationName: string;
   let webhookSubscriptionArgs: string;
   switch (deliveryMethod) {
-    case DeliveryMethod.Http || !versionSupportsEndpointField():
+    case DeliveryMethod.Http:
       mutationName = webhookId ? 'webhookSubscriptionUpdate' : 'webhookSubscriptionCreate';
       webhookSubscriptionArgs = `{callbackUrl: "${address}"}`;
       break;
--- a/src/webhooks/registry.ts
+++ b/src/webhooks/registry.ts
@@ -79,17 +79,9 @@ function isSuccess(result: any, deliveryMethod: DeliveryMethod, webhookId?: stri
   }
 }
 
-function versionSupportsEndpointField(
-  version: ApiVersion = Context.API_VERSION,
-) {
-  // 2020-07 onwards
-  return [
-    ApiVersion.July20,
-    ApiVersion.October20,
-    ApiVersion.January21,
-    ApiVersion.Unstable,
-    ApiVersion.Unversioned,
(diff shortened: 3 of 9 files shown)


src/base_types.ts
```diff
@@ -12,6 +12,7 @@ export interface ContextParams {
   SESSION_STORAGE?: SessionStorage;
   LOG_FILE?: string;
   USER_AGENT_PREFIX?: string;
+  EVENTBRIDGE_ARN: string;
```
paulomarg, line 15: I'm not sure we want to have this parameter in `Context`, since it's more of a global default than a library config. That said, making this a static setting in `Shopify.Webhooks.Registry` seems less than ideal. I'm torn 😄 

How much of a problem do you think it would be to simply not have a default for this? As far as I can see, apps will usually register webhooks in a single place after OAuth completes so it would be easy to just put the address in a variable and use it for all webhooks.

src/webhooks/types.ts
```diff
@@ -9,15 +9,22 @@ type WebhookHandlerFunction = (
   body: string
 ) => Promise<void>;
 
-export interface RegisterOptions {
+export type RegisterOptions = {
```
paulomarg, line 12: I really like the way you set up these types, but I'd consider putting the common fields into a private `RegisterBaseOptions` type so we can more easily evolve this in the future.

src/webhooks/types.ts
```diff
@@ -33,7 +40,13 @@ export interface WebhookRegistryEntry {
 interface WebhookCheckResponseNode {
   node: {
     id: string;
-    callbackUrl: string;
+    endpoint: {
```
paulomarg, line 43: Just noting that this change probably means we'd need a major release since after this since it'll break existing apps.

paulomarg (commented):
Thank you so much for this @rdunk! I really like the way you handled it, thought I'd jump in with some early thoughts, if it helps.

src/base_types.ts
rdunk: To be honest I wasn't sure either! I agree with the idea of not "polluting" the global context with variables that aren't used globally. And as you say, registering webhooks seems like it would happen in limited situations anyway (I'm no expert). Definitely happy to remove it from context if you think that's sensible.

src/webhooks/types.ts
rdunk: Yes, completely agree! I had the exact same thought about using an extendable base type for options but it was about 2am when I wrote that so I took the lazy route.

src/webhooks/types.ts
rdunk: Yeah, this is where I had some reservations. I don't think it would be too much effort to write some conditional logic to check against a given API version if you think it's worth doing? I noticed this library is backwards compatible for quite a few versions, so it seems like it would be a shame to drop that just for this feature.

src/webhooks/types.ts
rdunk: Alternatively, if we remove `EVENTBRIDGE_ARN` from `Context`, path would no longer be optional when `deliveryMethod` is of type `DeliveryMethod.EventBridge`. So this could be reverted to the previous interface with an extra `deliveryMethod` property of type `DeliveryMethod` anyway.

src/webhooks/registry.ts
```diff
@@ -144,8 +152,10 @@ const WebhooksRegistry: RegistryInterface = {
     webhookHandler,
   }: RegisterOptions): Promise<RegisterReturn> {
     const client = new GraphqlClient(shop, accessToken);
-    const address = `https://${Context.HOST_NAME}${path}`;
-
+    const address =
+      deliveryMethod && deliveryMethod === DeliveryMethod.EventBridge
```
paulomarg, line 156: We don't really need to check if `deliveryMethod` is set here since it defaults to `Http`.

paulomarg (commented):
Overall this looks good, though I haven't had a chance to test it yet!

The only consideration I have is that we may not be able to release this for a bit, since it will be a breaking change.

src/webhooks/registry.ts
rdunk: Good point, removed.

rdunk:
I've added backwards compatibility for API versions that don't support the `endpoint` field. These requests will revert to using the `node` level `callbackUrl` field instead. Any thoughts?

src/webhooks/registry.ts
```diff
@@ -78,8 +79,40 @@ function isSuccess(result: any, deliveryMethod: DeliveryMethod, webhookId?: stri
   }
 }
 
+function versionSupportsEndpointField(
```
paulomarg, line 82: I wonder if we should just break `version` apart and check if `year > 2020 || month >= 07`, so we don't have to remember to add future versions here (even though the old check will be deprecated fairly soon).

src/webhooks/registry.ts
```diff
@@ -107,11 +142,14 @@ function buildQuery(
...
   let webhookSubscriptionArgs: string;
   switch (deliveryMethod) {
-    case DeliveryMethod.Http:
+    case DeliveryMethod.Http || !versionSupportsEndpointField():
       mutationName = webhookId ? 'webhookSubscriptionUpdate' : 'webhookSubscriptionCreate';
       webhookSubscriptionArgs = `{callbackUrl: "${address}"}`;
       break;
     case DeliveryMethod.EventBridge:
+      if (!versionSupportsEndpointField()) {
+        throw new ShopifyErrors.UnsupportedClientType();
```
paulomarg, line 151: Nit: let's add an error message here to help make the reason for the error clearer.

paulomarg (commented):
I just realized that this is not a breaking change at all, since it only affects the internal check for existing webhooks, my mistake there!

This is pretty great, we'll test this and include it in the next release.

src/webhooks/registry.ts
```diff
@@ -107,11 +142,14 @@ function buildQuery(
   let mutationName: string;
   let webhookSubscriptionArgs: string;
   switch (deliveryMethod) {
-    case DeliveryMethod.Http:
+    case DeliveryMethod.Http || !versionSupportsEndpointField():
```
mkevinosullivan, line 145: If the version is pre-`2020-07` and `deliveryMethod` is passed in with `DeliveryMethod.EventBridge`, the result will be a GraphQL query built for `Http`, i.e., with the mutation being `webhookSubscriptionUpdate` or `webhookSubscriptionCreate` and `callbackUrl` being the `arn` passed in, though it should be (for a Http type query), a portion of a URL that is internally prefixed with `https://${Context.HOST_NAME}`.

This will result in a malformed query.

I think the check
```
if (deliveryMethod === DeliveryMethod.EventBridge && !versionSupportsEndpointField()) {
  // throw an error
}
```
should be added just after the method invocation (i.e., throw out the mismatched scenario early)

mkevinosullivan (changes requested):

src/webhooks/registry.ts
rdunk: Oh yes, good catch. I've added that in 22e510f25622dd6cf8c5fda05690b89efee79383.

src/webhooks/registry.ts
rdunk: I've added `EventBridge webhooks are not supported in API version "${Context.API_VERSION}".` as the message, combined with the change suggested by @mkevinosullivan.

src/webhooks/registry.ts
rdunk: I've added a new utility method (named `versionCompatible`  for now) as you suggested. I wasn't sure how we wanted to handle unstable/unversioned so those always return `true`.

src/utils/version-compatible.ts
```diff
@@ -0,0 +1,17 @@
+import {Context} from '../context';
+import {ApiVersion} from '../base_types';
+
+/**
+ * Check if the current or optionally supplied version is compatible with a given version
+ */
+export default function versionCompatible(referenceVersion: ApiVersion, currentVersion: ApiVersion = Context.API_VERSION): boolean {
```
paulomarg, line 7: This function will be super helpful in the future, thanks for adding it!

src/utils/version-compatible.ts
```diff
@@ -0,0 +1,17 @@
...
+  // Return true if not using a dated version
+  if (currentVersion === ApiVersion.Unstable || currentVersion === ApiVersion.Unversioned) {
+    return true;
+  }
+  const numericVersion = (version: string) => parseInt(version.replace('-', ''), 10);
+  const current = numericVersion(currentVersion);
+  const reference = numericVersion(referenceVersion);
+  console.log(current, reference, current >= reference);
```
paulomarg: Seems like this was left behind :)

src/webhooks/registry.ts
```diff
@@ -147,9 +143,6 @@ function buildQuery(
       webhookSubscriptionArgs = `{callbackUrl: "${address}"}`;
       break;
     case DeliveryMethod.EventBridge:
```
paulomarg: This case is OK, but I think the previous one could be changed:
```ts
case DeliveryMethod.Http || !versionSupportsEndpointField():
```

Since `Http` will always have a value, the expression will always evaluate to `DeliveryMethod.Http`, and the version check never runs (which is OK since we're doing the version check earlier now!).

src/webhooks/registry.ts
```diff
@@ -132,6 +124,10 @@ function buildQuery(
   deliveryMethod: DeliveryMethod = DeliveryMethod.Http,
   webhookId?: string,
 ): string {
+  if (deliveryMethod === DeliveryMethod.EventBridge && !versionSupportsEndpointField()) {
+    // No endpoint fallback exists for EventBridge, so throw early
+    throw new ShopifyErrors.UnsupportedClientType(`EventBridge webhooks are not supported in API version "${Context.API_VERSION}".`);
+  }
```
paulomarg, line 130: We may want to run this in `buildCheckQuery` as well, since that will be called before this. We could actually just run this as a gateway check in `register` to fail as early as we can.

paulomarg (commented):
Suggested a few tweaks, but I think we're close to a great solution now.

src/utils/version-compatible.ts
rdunk: I did indeed! That should have been fixed already in b87027b25a0855d59415a6ad5bf517b971a84eec.

src/webhooks/registry.ts
rdunk: Good point, I'll remove that.

src/webhooks/registry.ts
rdunk: I wasn't quite sure what to do here. I originally moved the check in `register` as you suggested, but realised that the `buildQuery` method is exported, so could potentially be consumed and run elsewhere without error being thrown (e.g. in tests), I guess the same is true of `buildCheckQuery`. What would you suggest? Do you think it just being in `register` is sufficient?

src/utils/version-compatible.ts
rdunk: Great!

src/webhooks/registry.ts
paulomarg: Hm, good catch - IMHO if the methods are exported we might as well go for the safer choice and just run the check on all exported methods to make sure we're covering all cases, it's cheaper to run it multiple times than it is to make a network request and wait for it to fail.

rdunk: fb6e23d32012d830b935a1ca961f901b96ae43a5 adds a common `validateDeliveryMethod` function which is executed on `register` and `buildQuery` as both those function have access to the `deliveryMethod`.

I suppose `buildCheckQuery` is perhaps fine without the check as it doesn't actually care which delivery method you are using and only accepts a topic as a parameter? For instance there may be a use case where you would want to query for an already registered webhook without knowing what type of delivery method that webhook is actually utilising.

rdunk:
Not so experienced with GH reviews, but I think all of the requested changes have been implemented now?

paulomarg:
Hey @rdunk, sorry for the delay in responding - I think all of the changes look good now, I just haven't had a chance to test it out myself before merging, but I was actually working on that yesterday. I'll keep you posted!

paulomarg:
I tested this myself, and I was able to register both legacy and current HTTP webhooks, and the library code worked for ARN endpoints too (even though I didn't have a valid one to test, the request we fired was correct).

Huge thanks to @rdunk for your amazing contribution and patience!

rdunk:
@paulomarg Great stuff! No worries about the delay at all, thanks very much for the thorough reviews and input. Hopefully others find this useful too.

PR #128, opened 2021-03-08
[Proposal] Enable library to query Storefront API

paulomarg:

We're considering adding support for the Storefront API to the GraphQL client. This is a prototype of what that might look like.

Rearranging the constructor params for the GraphQL client so that it allows an optional `apiType` setting (which defaults to `Admin`), and changing the client to change the request endpoint accordingly. For consistency, the REST client was also updated so its constructor params work in the same way, even though there is no REST Storefront API.

The constructors were updated in a way that doesn't break existing apps, and we can easily change them when we release a major version (or when we need to add a mandatory param to the client).

An app's code might look like this to make Storefront requests:

```ts
        const client = new Shopify.Clients.Rest({
          domain: myShop,
          accessToken: myToken,
        });
        const sfTokenResponse = await client.post({
          path: 'storefront_access_tokens',
          type: DataType.JSON,
          data: {
            "storefront_access_token": {
              "title": "Test"
            }
          },
        });

        const sfClient = new Shopify.Clients.Graphql({
          apiType: ApiClientType.Storefront,
          domain: myShop,
          accessToken: sfTokenResponse.body['storefront_access_token']['access_token'],
        });
        const sfGqlResponse = await sfClient.query({
          data: `{
            shop {
              name
              primaryDomain {
                url
                host
              }
            }
          }`,
        });
```

## Type of change

## Checklist

### DIFF
--- a/src/base_types.ts
+++ b/src/base_types.ts
@@ -12,6 +12,7 @@ export interface ContextParams {
   SESSION_STORAGE?: SessionStorage;
   LOG_FILE?: string;
   USER_AGENT_PREFIX?: string;
+  PRIVATE_APP_STOREFRONT_ACCESS_TOKEN?: string;
 }
 
 export enum ApiVersion {
@@ -29,6 +30,7 @@ export enum ApiVersion {
 
 export enum ShopifyHeader {
   AccessToken = 'X-Shopify-Access-Token',
+  StorefrontAccessToken = 'X-Shopify-Storefront-Access-Token',
   Hmac = 'X-Shopify-Hmac-Sha256',
   Topic = 'X-Shopify-Topic',
   Domain = 'X-Shopify-Shop-Domain',
--- a/src/clients/graphql/graphql_client.ts
+++ b/src/clients/graphql/graphql_client.ts
@@ -2,20 +2,43 @@ import {MissingRequiredArgument} from '../../error';
 import {Context} from '../../context';
 import {ShopifyHeader} from '../../base_types';
 import {HttpClient} from '../http_client/http_client';
-import {DataType, RequestReturn} from '../http_client/types';
+import {ApiClientParams, ApiClientType, DataType, RequestReturn} from '../http_client/types';
 import * as ShopifyErrors from '../../error';
 
 import {GraphqlParams} from './types';
 
+interface AccessTokenHeader {
+  header: string;
+  value: string;
+}
+
 export class GraphqlClient {
+  readonly apiType: ApiClientType;
+  readonly domain: string;
+  readonly accessToken?: string;
+
   private readonly client: HttpClient;
 
-  constructor(readonly domain: string, readonly token?: string) {
-    if (!Context.IS_PRIVATE_APP && !token) {
-      throw new ShopifyErrors.MissingRequiredArgument('Missing access token when creating GraphQL client');
+  // When we next release a major version for this library, we should remove string as a valid type for params and
+  // remove the accessToken param. Also remove the first block.
+  public constructor(params: ApiClientParams | string, accessToken?: string) {
+    if (typeof params === 'string') {
+      // eslint-disable-next-line no-param-reassign
+      params = {
+        domain: params,
+        accessToken,
+      };
     }
 
+    this.domain = params.domain;
     this.client = new HttpClient(this.domain);
+
+    this.apiType = params.apiType || ApiClientType.Admin;
+    this.accessToken = params.accessToken;
+
+    if (!Context.IS_PRIVATE_APP && !this.accessToken) {
+      throw new ShopifyErrors.MissingRequiredArgument('Missing access token when creating GraphQL client');
+    }
   }
 
   async query(params: GraphqlParams): Promise<RequestReturn> {
@@ -23,11 +46,12 @@ export class GraphqlClient {
       throw new MissingRequiredArgument('Query missing.');
     }
 
+    const accessTokenHeader: AccessTokenHeader = this.getAccessTokenHeaderForApiType();
     params.extraHeaders = {
-      [ShopifyHeader.AccessToken]: Context.IS_PRIVATE_APP ? Context.API_SECRET_KEY : this.token as string,
+      [accessTokenHeader.header]: accessTokenHeader.value,
       ...params.extraHeaders,
     };
-    const path = `/admin/api/${Context.API_VERSION}/graphql.json`;
+    const path = `${this.getBasePathForApiType()}/${Context.API_VERSION}/graphql.json`;
 
     let dataType: DataType.GraphQL | DataType.JSON;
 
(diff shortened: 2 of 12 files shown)


thecodepixi:
This is really interesting. It seems (like you mentioned in the PHP tech doc) like an easy win to add this feature to the library. It's doesn't add much complexity and would be useful to partners. It also seems like there's probably an opportunity to add a convenience/util method that takes care of the REST request for the SF token? Just a thought 🤔

src/base_types.ts
```diff
@@ -12,6 +12,7 @@ export interface ContextParams {
   SESSION_STORAGE?: SessionStorage;
   LOG_FILE?: string;
   USER_AGENT_PREFIX?: string;
+  PRIVATE_APP_STOREFRONT_ACCESS_TOKEN?: string;
```
thecodepixi, line 15: Slightly confused by this naming. Do only private apps use the SF API?

andyw8:
Not sure how feasible this is, but should users need to specify the `apiType`? Could it be determined programatically based on the request?

src/base_types.ts
paulomarg: There are 2 ways to get access tokens: 
1. Via your admin, where you grant a private app SF permission and get a hard-coded token (which would go into this field)
2. The example in the PR description, where you can create public tokens for a sales channel

I wanted to make it very clear that this field was meant only for private apps, which is why the name is quite the mouthful.

paulomarg:
> Not sure how feasible this is, but should users need to specify the `apiType`? Could it be determined programatically based on the request?

I don't think we can detect that - the SF API is GraphQL only, so we'd probably need to parse the actual query to be able to tell where it would land, and even so there are overlaps in what's available.

Basically, the only constant difference between the two APIs is the endpoint (`/admin/api` vs `/api`), so we'd either need to take that in or a type param. I went with the type to make it easier to add other API types in the future if that is ever needed, and so we have a simple way to check which API we're querying for e.g. the access token headers.

carmelal:
> For consistency, the REST client was also updated so its constructor params work in the same way, even though there is no REST Storefront API.
Could that potentially be extra confusing? Consistency is good, but it may create the illusion of a choice that isn't actually there and give the user more opportunity to do something they shouldn't be doing (attempting to reach REST API).

thecodepixi:
> > > For consistency, the REST client was also updated so its constructor params work in the same way, even though there is no REST Storefront API.
> > 
> > 
> > Could that potentially be extra confusing? Consistency is good, but it may create the illusion of a choice that isn't actually there and give the user more opportunity to do something they shouldn't be doing (attempting to reach REST API).
> 
> That is true, and I added code to make sure we don't allow invalid types here, but I wonder if that is confusing. I'd love hear what everyone thinks.

I think I agree with Carmela here that this may cause confusion/illusion of choice. I don't think the two need to be consistent in this way if it makes the REST usage/capability less clear.

paulomarg:
Thanks for the great points everyone raised, they make a lot of sense. Since Storefront doesn't even allow REST, I figured it would actually be better to just extend the GraphQL client and abstract a few of its parts.

The code to use this would now look like
```ts

      const client = new Shopify.Clients.Rest({
        domain: session.shop,
        accessToken: session.accessToken,
      });
      const sfTokenResponse = await client.post({
        path: 'storefront_access_tokens',
        type: DataType.JSON,
        data: {
          "storefront_access_token": {
            "title": "Test"
          }
        },
      });

      const sfClient = new Shopify.Clients.Storefront({
        domain: session.shop,
        accessToken: sfTokenResponse.body['storefront_access_token']['access_token'],
      });
      const sfGqlResponse = await sfClient.query({
        data: `...`,
      });
```

src/clients/graphql/storefront_client.ts
```diff
@@ -0,0 +1,17 @@
...
+import {GraphqlClient, AccessTokenHeader} from './graphql_client';
+
+export class StorefrontClient extends GraphqlClient {
+  protected baseApiPath = '/api';
+
+  protected getAccessTokenHeader(): AccessTokenHeader {
+    return {
+      header: ShopifyHeader.StorefrontAccessToken,
+      value: (
+        Context.IS_PRIVATE_APP ? Context.PRIVATE_APP_STOREFRONT_ACCESS_TOKEN || this.accessToken : this.accessToken
```
thecodepixi, line 13: I'm not sure I understand the `||` in the first condition on this line. Can you explain what's happening here?

paulomarg: We're basically defaulting to the given access token if the setting isn't there. Not sure we necessarily want to do that, though - we don't in the GraphQL client, but then again it uses the secret key which is mandatory.

mkevinosullivan: Does
```
Context.PRIVATE_APP_STOREFRONT_ACCESS_TOKEN || this.accessToken
```
evaluate to a string or a boolean?

src/clients/graphql/test/storefront_client.test.ts
```diff
@@ -0,0 +1,77 @@
+import '../../../test/test_helper';
+import {ShopifyHeader} from '../../../base_types';
+import {assertHttpRequest} from '../../http_client/test/test_helper';
+import {StorefrontClient} from '../storefront_client';
+import {Context} from '../../../context';
+
+const DOMAIN = 'shop.myshopify.com';
```
mkevinosullivan, line 7: Should we make the test `DOMAIN` end with `.io`, i.e., `shop.myshopify.io`?

src/base_types.ts
mkevinosullivan: How do you obtain those public tokens for a sales channel? and how would they be used in an app?  Are they retrieved via an API or does the app developer need to load them from the environment/database/files/etc.?

src/clients/graphql/graphql_client.ts
```diff
@@ -2,32 +2,57 @@ import {MissingRequiredArgument} from '../../error';
...
+  protected baseApiPath = '/admin/api';
+
   private readonly client: HttpClient;
 
-  constructor(readonly domain: string, readonly token?: string) {
-    if (!Context.IS_PRIVATE_APP && !token) {
-      throw new ShopifyErrors.MissingRequiredArgument('Missing access token when creating GraphQL client');
+  // When we next release a major version for this library, we should remove string as a valid type for params and
+  // remove the accessToken param. Also remove the first block.
+  public constructor(params: ApiClientParams | string, accessToken?: string) {
```
mkevinosullivan, line 25: so the actual proposed constructor will be?
```
public constructor(params: ApiClientParams) { ... }
```
Since this is a proposal, why not just change it across the library?

mkevinosullivan: Actually, why change it to a type?
From, essentially ...
```
constructor(domain: string, token?: string)
```
...to
```
constructor({domain: string, token?: string})
```

src/base_types.ts
paulomarg: Public apps will use the second path above, by querying the `storefront_access_token` endpoint the Admin API. The code I used in the example is for public apps.

src/clients/graphql/graphql_client.ts
paulomarg: Yes, this isn't strictly necessary, but in most other places in the library we're using types rather than separate parameters (which makes functions easier to evolve), so I was going for consistency.

The reason I didn't make the full change is to keep this backward compatible until we need to do a major release. Current apps would be relying on the old format and I was trying to avoid breaking them if we could.

All of that said, I have no strong feelings about doing this. This is a fairly simple refactoring that we can do at any time.

src/clients/graphql/storefront_client.ts
paulomarg: String:
```
a = 'a';
"a"
b = 'b';
"b"
a || b
"a"
```

src/clients/graphql/graphql_client.ts
paulomarg: I'm rolling back this change, it should be done in a separate PR if we're not making any actual changes to constructors here any more.

thecodepixi (approved):
I'm on board with this as it stands. Do we want to add documentation updates so it can all be released at the same time? Also wondering if we need to do anything in terms of announcing that SF API support will be/is available. 🤔

paulomarg:
> I'm on board with this as it stands. Do we want to add documentation updates so it can all be released at the same time? Also wondering if we need to do anything in terms of announcing that SF API support will be/is available. 🤔

Yes, I've updated the docs now, thanks for the reminder. I've also changed mentions of `Shopify Admin API Library` to `Shopify API Library`.

docs/usage/storefront.md
```diff
@@ -0,0 +1,38 @@
+# Make a Storefront API call
+
+The library also allows you to send GraphQL requests to the [Shopify Storefront API](https://shopify.dev/docs/storefront-api). To do that, you can use the `StorefrontClient` using the current shop URL and an `accessToken`.
+
+You can obtain Storefront API access tokens for both private apps and sales channels. Please read [our documentation](https://shopify.dev/docs/storefront-api/getting-started) to learn more about Storefront Access Tokens.
```
thecodepixi, line 5: Do we want to also provide an example like you have in the PR comment showing how to request this information from the REST API? Or do we feel like directing them to the docs is enough? I think giving the example post request for the token might be useful to keep everything in the same place and show explicitly how to make that request using the library clients.

paulomarg: Great point, I think that's helpful. I've added the example for the PR there.

mkevinosullivan (approved):
LGTM 🚀

thecodepixi (approved):
This continues to LGTM 👍

paulomarg:
> This continues to LGTM 👍

Sorry, I accidentally the review button for you 😬

PR #131, opened 2021-03-10
Throw specific error for missing cookies on OAuth

paulomarg:

As mentioned on https://github.com/Shopify/koa-shopify-auth/issues/61#issuecomment-794850778, we were throwing the same error (`SessionNotFound`) if the session itself was missing, or if the cookie the holds its id was missing, which can be confusing when errors happen.

Throwing a specific `CookieNotFound` error if we can't load the session id from the cookie to make it easier to surface issues to apps.

## Type of change

## Checklist


thecodepixi (approved):
Makes sense to me!

pratiknikam:
@paulomarg thanks for this! When should we expect new release tag for `@shopify/koa-shopify-auth`?

PR #132, opened 2021-03-12
Ensure session expirations are Dates

paulomarg:

It was pointed out (https://github.com/Shopify/koa-shopify-auth/issues/65#issuecomment-797591927) that the `CustomSessionStorage` class may cause issues if the returned session `expires` field isn't a `Date` object.

Making sure that we allow strings to be returned for the expiration date, and handling the conversion to `Date`.

## Type of change

## Checklist

### DIFF
--- a/src/auth/session/storage/custom.ts
+++ b/src/auth/session/storage/custom.ts
@@ -34,10 +34,19 @@ export class CustomSessionStorage implements SessionStorage {
     }
     if (result) {
       if (result instanceof Session) {
+        if (result.expires && typeof result.expires === 'string') {
+          result.expires = new Date(result.expires);
+        }
+
         return result;
       } else if (result instanceof Object && 'id' in result) {
         let session = new Session(result.id as string);
         session = {...session, ...result};
+
+        if (session.expires && typeof session.expires === 'string') {
+          session.expires = new Date(session.expires);
+        }
+
         return session;
       } else {
         throw new ShopifyErrors.SessionStorageError(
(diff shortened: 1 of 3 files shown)


src/auth/session/storage/custom.ts
```diff
@@ -34,10 +34,19 @@ export class CustomSessionStorage implements SessionStorage {
...
+        }
+
         return result;
       } else if (result instanceof Object && 'id' in result) {
         let session = new Session(result.id as string);
         session = {...session, ...result};
+
+        if (session.expires && typeof session.expires === 'string') {
+          session.expires = new Date(session.expires);
+        }
```
kato-takaomi-ams, line 48:
```suggestion
        const session = Session.cloneSession(result, result.id);
```

and following part port to Session.cloneSession.
```ts
        if (session.expires && typeof session.expires === 'string') {
          session.expires = new Date(session.expires);
        }
```

paulomarg: Hmm, I don't think that would work - we can't guarantee that `result` will be of type `Session` here, since it may have been loaded from something like a `JSON.parse()` call.

If we tried to call `Session.clone` with `result` it would break in Typescript, and we shouldn't expect the `expires` field to be a string there (it is typed as `Date` so it would have to be set when the session is created anyway).

thecodepixi (approved):
LGTM! Thanks for fixing this!

src/auth/session/storage/custom.ts
kato-takaomi-ams: @paulomarg I got it.
My suggestion is for the following reasons:

I wish I could improve the point that users have to convert the `session.expires` to Date type.
(This seems to be solved if `session.isActive()` is implemented)

`session = {...session, ...result};` overwrites `session.id` with `result.id`.
If this assignment is OK, why does `Session.cloneSession` exist?

PR #135, opened 2021-03-13
Fix get request with query

### LINKED ISSUE #134: Rest client. Get request with query fails if the limit is not specified
# Issue summary
With this kind of request
```
client.get({
  path: 'redirects',
  query: {
	  "path": 'some_path',
  },
});
```

I get error, because limit is undefined

## Expected behavior

Successful request to the api

## Actual behavior

I get error UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'toString' of undefined
which comes from this row https://github.com/Shopify/shopify-node-api/blob/7c9201af556c6f82deb936215f5099d08e35f727/src/clients/rest/rest_client.ts#L35

---

## Checklist

derom:

Fixes #134

Added test to catch the described error. Made limit optional, because it has default value (50)

## Type of change

## Checklist

### DIFF
--- a/src/clients/rest/rest_client.ts
+++ b/src/clients/rest/rest_client.ts
@@ -31,10 +31,15 @@ class RestClient extends HttpClient {
 
     const link = ret.headers.get('link');
     if (params.query && link !== undefined) {
-      const pageInfo: PageInfo = {};
+      const DEFAULT_LIMIT = '50';
+      let limit = DEFAULT_LIMIT;
       if (params.query.limit) {
-        pageInfo.limit = params.query.limit.toString();
+        limit = params.query.limit.toString();
       }
+
+      const pageInfo: PageInfo = {
+        limit: limit,
+      };
 
       if (link) {
         const links = link.split(', ');
--- a/src/clients/rest/rest_client.ts
+++ b/src/clients/rest/rest_client.ts
@@ -32,13 +32,8 @@ class RestClient extends HttpClient {
 
     const link = ret.headers.get('link');
     if (params.query && link !== undefined) {
-      let limit = RestClient.DEFAULT_LIMIT;
-      if (params.query.limit) {
-        limit = params.query.limit.toString();
-      }
-
       const pageInfo: PageInfo = {
-        limit,
+        limit: params.query.limit ? params.query.limit.toString() : RestClient.DEFAULT_LIMIT,
       };
 
       if (link) {
--- a/src/clients/rest/rest_client.ts
+++ b/src/clients/rest/rest_client.ts
@@ -31,14 +31,14 @@ class RestClient extends HttpClient {
 
     const link = ret.headers.get('link');
     if (params.query && link !== undefined) {
-      const DEFAULT_LIMIT = '50';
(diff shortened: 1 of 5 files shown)


thecodepixi (approved):
This makes sense to me. We should definitely be able to make REST requests without passing a limit, since the API will just apply its default limit to the request. Thanks for catching that.

derom:
Thanks for review @thecodepixi
I've pushed some changes. ```limit``` in PageInfo is non-optional again. I realized that the limit is always known and we should just set the default value if a user didn't specify the limit. Also, unused type PageInfoParams was removed

src/clients/rest/rest_client.ts
```diff
@@ -31,8 +31,14 @@ class RestClient extends HttpClient {
 
     const link = ret.headers.get('link');
     if (params.query && link !== undefined) {
+      const defaultLimit = '50';
```
thecodepixi, line 34: I might put this as a static variable on the class instead, as we have with the regex: 

```suggestion
     private static DEFAULT_LIMIT = '50';
```

and then we can still reference it with the same type of conditional statement.

thecodepixi:
Taking another look at this now, and I'm not sure we can entirely do away with `PageInfo, but have asked for additional opinions on that front.

paulomarg:
I agree, it's a bit strange that it doesn't cause any tests to break, but we're definitely still using that type in the REST client class.

derom:
@thecodepixi @paulomarg PageInfoParams was replaced by PageInfo in this commit https://github.com/Shopify/shopify-node-api/commit/53b5e0779e55b9e554aa88b1a8d0d34c4e655111
I didn't find any usages of that interface https://github.com/Shopify/shopify-node-api/search?q=PageInfoParams

src/clients/rest/rest_client.ts
derom: +

thecodepixi:
Ah OK, maybe that was my misunderstanding. You're proposing just removing `PageInfoParams` but not `PageInfo`?

derom:
Yes, only PageInfoParams

src/clients/rest/rest_client.ts
```diff
@@ -31,8 +32,13 @@ class RestClient extends HttpClient {
 
     const link = ret.headers.get('link');
     if (params.query && link !== undefined) {
+      let limit = RestClient.DEFAULT_LIMIT;
```
thecodepixi, line 35: Nit: This feels a bit redundant. Can we accomplish this without the reassignment on this line?

thecodepixi:
> Yes, only PageInfoParams

Oh then yeah, if that's no longer needed I would think it's fine to remove.

src/clients/rest/rest_client.ts
derom: Agree. I think it looks better now

paulomarg (approved):
I like it too! 🚀

derom:
Should I do anything else to merge this pr?

thecodepixi:
> Should I do anything else to merge this pr?

I gotchu!

PR #136, opened 2021-03-15
Rename docs/index.md to docs/README.md

gfscott:

Since the primary context for consuming this content is directly browsing GitHub, if we change this filename to README.md, it will display the page content when browsing this directory. That doesn't happen for `index.md` files, so you'd need to click through to actually see the table of contents.

## Type of change

## Checklist

- [ ] ~I have added/updated tests for this change~


paulomarg (commented):
I have no problems with this, but we may need to change the `/README.md` file to point here instead of `index`.

thecodepixi (approved):
You beat me to it! I just noticed this while working on my other docs PR. Ditto Paulo's comment about changing the link in the main README though.

paulomarg (approved):
🚢

PR #153, opened 2021-03-30
Add isActive() session method

### LINKED ISSUE #133: port verifyToken from koa-shopify-auth
[verify-tokens.ts](https://github.com/Shopify/koa-shopify-auth/blob/b3a644bf7955a658c2b5c0347e73170934bdd4e6/src/verify-request/verify-token.ts#L13)

I want to use it like this.
```ts
if (!Shopify.Auth.verifyToken(req, res, "online") {
  return res.redirect(`/auth?${req.query.shop}`);
}
```

carmelal:
### Why are these changes introduced?
* Fixes #133 
* Easier to see whether a session is active

### What is this pull request doing?
* Adds method checking if a session is active
* Adds `SessionInterface` type to use instead of `Session`
* Moves session storage tests to correct directory level
* Layers `auth` exports

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 ## Unreleased
 ### Added
 - Added Storefront API client under `Shopify.Clients.Storefront`
+- Add `isActive()` method to `Session` class to check if session is active, replace `Session` with `SessionInterface` when used as a type [#153](https://github.com/Shopify/shopify-node-api/pull/153)
 
 ## [1.2.1] - 2021-03-26
 ### Added
--- a/src/auth/session/session.ts
+++ b/src/auth/session/session.ts
@@ -1,9 +1,12 @@
 import {OnlineAccessInfo} from '../oauth/types';
+import {Context} from '../../context';
+
+import {SessionInterface} from './types';
 
 /**
  * Stores App information from logged in merchants so they can make authenticated requests to the Admin API.
  */
-class Session {
+class Session implements SessionInterface {
   public static cloneSession(session: Session, newId: string): Session {
     const newSession = new Session(newId);
 
@@ -27,6 +30,14 @@ class Session {
   public onlineAccessInfo?: OnlineAccessInfo;
 
   constructor(readonly id: string) {}
+
+  public isActive(): boolean {
+    const scopesUnchanged = Context.SCOPES.equals(this.scope);
+    if (scopesUnchanged && this.accessToken && (!this.expires || this.expires >= new Date())) {
+      return true;
+    }
+    return false;
+  }
 }
 
 export {Session};
--- a/src/auth/session/storage/custom.ts
+++ b/src/auth/session/storage/custom.ts
@@ -1,11 +1,12 @@
 import {Session} from '../session';
+import {SessionInterface} from '../types';
 import {SessionStorage} from '../session_storage';
 import * as ShopifyErrors from '../../../error';
 
 export class CustomSessionStorage implements SessionStorage {
   constructor(
-    readonly storeCallback: (session: Session) => Promise<boolean>,
-    readonly loadCallback: (id: string) => Promise<Session | Record<string, unknown> | undefined>,
+    readonly storeCallback: (session: SessionInterface) => Promise<boolean>,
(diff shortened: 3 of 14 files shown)


src/auth/session/test/session.test.ts
```diff
@@ -0,0 +1,50 @@
+import {ApiVersion, ContextParams} from '../../../base_types';
+import {Context} from '../../../context';
+import '../../../test/test_helper';
+import {Session} from '..';
+
+const validContextParams: ContextParams = {
```
thecodepixi, line 6: Valid `Context` is initialized in `text_helper`. Is there a reason we need to create a new `Context` for these tests?

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 ## [1.2.1] - 2021-03-26
 ### Added
 - Added `April21` to `ApiVersion` [#149](https://github.com/Shopify/shopify-node-api/pull/149)
+- Added `isActive()` method to `Session` class to check if session is active [#153](https://github.com/Shopify/shopify-node-api/pull/153)
```
thecodepixi, line 13: I might add to the changelog that certain things will now use `SessionInterface` instead of just `Session`, particularly where `Session` used to be the expected return type or argument type.

src/auth/session/test/session.test.ts
paulomarg: Not impossible that this was added before the test_helper case 🤔

paulomarg (approved):
LGTM, I'm happy for this to go in after we deal with my comment.

src/auth/session/session.ts
```diff
@@ -27,6 +28,14 @@ class Session {
   public onlineAccessInfo?: OnlineAccessInfo;
 
   constructor(readonly id: string) {}
+
+  public isActive(): boolean {
+    const scopesChanged = !Context.SCOPES.equals(this.scope);
+    if (!scopesChanged && this.accessToken && (!this.expires || this.expires >= new Date())) {
```
mkevinosullivan, line 34:
```suggestion
    const scopesUnchanged = Context.SCOPES.equals(this.scope);
    if (scopesUnchanged && this.accessToken && (!this.expires || this.expires >= new Date())) {
```

mkevinosullivan: Wondering if removing the two negatives might make this a little easier to read

mkevinosullivan: Shouldn't `class Session` implement `SessionInterface`?

src/auth/session/storage/custom.ts
```diff
@@ -1,19 +1,20 @@
...
+import {SessionInterface} from '../types';
 import {SessionStorage} from '../session_storage';
 import * as ShopifyErrors from '../../../error';
 
 export class CustomSessionStorage implements SessionStorage {
   constructor(
-    readonly storeCallback: (session: Session) => Promise<boolean>,
-    readonly loadCallback: (id: string) => Promise<Session | Record<string, unknown> | undefined>,
+    readonly storeCallback: (session: SessionInterface) => Promise<boolean>,
+    readonly loadCallback: (id: string) => Promise<SessionInterface | Record<string, unknown> | undefined>,
```
mkevinosullivan, line 9: Is `Record<string, unknown` a valid return type for the `loadCallback` method?  Wouldn't the two valid return types be either `SessionInterface` or `undefined`?

carmelal: [This](https://github.com/Shopify/shopify-node-api/pull/126) PR explains why this option was added.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 ## [1.2.1] - 2021-03-26
 ### Added
 - Added `April21` to `ApiVersion` [#149](https://github.com/Shopify/shopify-node-api/pull/149)
+- Add `isActive()` method to `Session` class to check if session is active, replace `Session` with `SessionInterface` when used as a type [#153](https://github.com/Shopify/shopify-node-api/pull/153)
```
thecodepixi, line 15: This should probably go under the `unreleased` heading!

thecodepixi (approved):
This looks great! I think you just need to move your `changelog` note up a bit, but otherwise approved!

PR #154, opened 2021-03-31
Add reference to list of topics for Shopify.Webhooks.Registry.register()

t-kelly:
I had to hunt this down by searching this repo and finding it in a comment - I think it would be helpful to surface it in our examples!


t-kelly:
I was looking at that originally but didn't see anything obvious. Will merge this for now and let you do a followup where you feel it could be done best!

PR #169, opened 2021-04-28
DRAFT Moves required Session args to constructor

thecodepixi:

Fixes #0000 

## Type of change

## Checklist


src/auth/scopes/index.ts
```diff
@@ -4,11 +4,11 @@ class AuthScopes {
   private compressedScopes: Set<string>;
   private expandedScopes: Set<string>;
 
-  constructor(scopes: string | string[]) {
+  constructor(scopes: string | string[] | undefined) {
     let scopesArray: string[] = [];
     if (typeof scopes === 'string') {
       scopesArray = scopes.split(new RegExp(`${AuthScopes.SCOPE_DELIMITER}\\s*`));
-    } else {
+    } else if (scopes) {
```
paulomarg, line 11: Should we be extra sure that we only process accepted types?

```suggestion
    } else if (scopes !== undefined) {
```

src/auth/session/session.ts
```diff
@@ -1,32 +1,32 @@
...
-  public isOnline?: boolean;
   public accessToken?: string;
   public onlineAccessInfo?: OnlineAccessInfo;
 
-  constructor(readonly id: string) {}
+  constructor(
+    readonly id: string,
+    public shop: string,
+    public state: string,
+    public isOnline: boolean = false,
```
paulomarg, line 28: I wonder if we want the default here - I'd prefer to always have the session be explicitly defined as online / offline.

paulomarg (commented):
Looks great, just had a couple of nits!

src/auth/scopes/index.ts
thecodepixi: That's a good point. Should there be an exception/error at that point if `scopes` is not a string, array or undefined?

src/auth/session/session.ts
thecodepixi: That's a good question. We've always defaulted to offline and I don't think ever revisited that decision? I'll see what breaks if we remove it!

src/auth/scopes/index.ts
paulomarg: Technically, typescript would fail in that case, but javascript wouldn't. Then again, we don't check for that sort of thing in a lot of other places (and is that an excuse for not doing it here? 😅 ).

thecodepixi:
This needs a good rebase and probably a bit of refactoring since we've merged #153. I'll get that taken care of so this should be in a good place soon.

CHANGELOG.md
```diff
@@ -10,9 +10,13 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 - Added Storefront API client under `Shopify.Clients.Storefront`
 - Add `isActive()` method to `Session` class to check if session is active, replace `Session` with `SessionInterface` when used as a type [#153](https://github.com/Shopify/shopify-node-api/pull/153)
 
+### Fixed
+- Required `Session` arguments must be passed to the constructor [#169](https://github.com/Shopify/shopify-node-api/pull/169)
+- Allow `undefined` in `AuthScopes` [#169](https://github.com/Shopify/shopify-node-api/pull/169)
 ## [1.2.1] - 2021-03-26
 ### Added
 - Added `April21` to `ApiVersion` [#149](https://github.com/Shopify/shopify-node-api/pull/149)
+- Sessions no longer default to `false` for `isOnline` [#169](https://github.com/Shopify/shopify-node-api/pull/169)
```
paulomarg, line 19: I'm a little confused - is this part of v1.2.1?

Also, I think this might be a breaking change, right? It will essentially break any apps that build sessions (which might be the case for custom session storages made prior to the change to allow objects).
**Update**: I actually don't think it's breaking! It's OK to not get those values in the constructor (if we make the params optional), and since the properties are all public, they can continue to be set as they were.

src/auth/scopes/index.ts
JaKXz: I think an easy way to handle something like this is using something like [`lodash.castArray`](https://lodash.com/docs/#castArray) so that we're always dealing with a list, so:

```jsx
// type any for illustration
constructor(scopes: any) {
  let scopesArray = castArray(scopes).filter(Boolean)
}
```

src/auth/session/session.ts
```diff
@@ -1,3 +1,4 @@
+// import {Context} from '../../context';
```
JaKXz, line 1: 🔥

CHANGELOG.md
thecodepixi: Ope, sorry. My changelog got all weird after I rebased. Will fix.

thecodepixi: Also yes, this would be breaking. I don't think these values _can_ be optional? They need to be there for other parts of the code that handle the `Session` to work, right?

paulomarg: Yes, but if you call the constructor and then set the fields, that's OK too, as long as the right fields continue to be set - i.e. apps that did it 'the old way' would continue to work if the constructor doesn't break because of missing params.

mariomx3:
Will this be merged in soon? Really critical

PR #181, opened 2021-05-14
Add support for PubSub webhook registration

tbrenev:

We will soon have support for webhook delivery via Google Cloud PubSub.
We need to update the webhook registration function in this library to support this new delivery method.

The `register` function already takes in a `DeliveryMethod`. This PR adds a new `DeliveryMethod.PubSub`,
and implements the appropriate behaviour for it. In particular, `register` hits the new `pubSubWebhookSubscriptionCreate`
and `pubSubWebhookSubscriptionUpdate` endpoints in the GraphQL API.

This PR also adds some documentation about registering non-HTTP-delivery webhooks (the docs for
EventBridge webhooks were missing as well).

## Type of change

## Checklist

### DIFF
--- a/docs/usage/webhooks.md
+++ b/docs/usage/webhooks.md
@@ -74,6 +74,20 @@ app.get('/auth/callback', async (req, res) => {
 ```
 </details>
 
+### EventBridge and PubSub Webhooks
+
+You can also register webhooks for delivery to Amazon EventBridge or Google Cloud
+Pub/Sub. In this case the `path` argument to
+`Shopify.Webhooks.Registry.register` needs to be of a specific form.
+
+For EventBridge, the `path` must be the [ARN of the partner event
+source](https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_EventSource.html).
+
+For Pub/Sub, the `path` must be of the form
+`pubsub://[PROJECT-ID]:[PUB-SUB-TOPIC-ID]`.  For example, if you created a topic
+with id `red` in the project `blue`, then the value of `path` would be
+`pubsub://red:blue`.
+
 ## Process a Webhook
 
 To process a webhook, you need to listen on the route(s) you provided during the Webhook registration process, then call the appropriate handler.  The library provides a convenient `process` method that acts as a middleware to handle webhooks. It takes care of calling the correct handler for the registered Webhook topics.
--- a/src/base_types.ts
+++ b/src/base_types.ts
@@ -25,6 +25,7 @@ export enum ApiVersion {
   October20 = '2020-10',
   January21 = '2021-01',
   April21 = '2021-04',
+  July21 = '2021-07',
   Unstable = 'unstable',
   Unversioned = 'unversioned',
 }
--- a/src/webhooks/registry.ts
+++ b/src/webhooks/registry.ts
@@ -45,38 +45,26 @@ interface RegistryInterface {
 }
 
 function isSuccess(result: any, deliveryMethod: DeliveryMethod, webhookId?: string): boolean {
+  let endpoint;
   switch (deliveryMethod) {
     case DeliveryMethod.Http:
-      if (webhookId) {
-        return Boolean(
-          result.data &&
-            result.data.webhookSubscriptionUpdate &&
-            result.data.webhookSubscriptionUpdate.webhookSubscription,
-        );
-      } else {
-        return Boolean(
-          result.data &&
-            result.data.webhookSubscriptionCreate &&
-            result.data.webhookSubscriptionCreate.webhookSubscription,
-        );
-      }
+      endpoint = 'webhookSubscription';
+      break;
     case DeliveryMethod.EventBridge:
-      if (webhookId) {
-        return Boolean(
-          result.data &&
-            result.data.eventBridgeWebhookSubscriptionUpdate &&
-            result.data.eventBridgeWebhookSubscriptionUpdate.webhookSubscription,
-        );
-      } else {
-        return Boolean(
-          result.data &&
-            result.data.eventBridgeWebhookSubscriptionCreate &&
-            result.data.eventBridgeWebhookSubscriptionCreate.webhookSubscription,
-        );
-      }
+      endpoint = 'eventBridgeWebhookSubscription';
+      break;
+    case DeliveryMethod.PubSub:
+      endpoint = 'pubSubWebhookSubscription';
+      break;
     default:
       return false;
   }
+  endpoint += webhookId ? 'Update' : 'Create';
+  return Boolean(
+    result.data &&
+      result.data[endpoint] &&
+      result.data[endpoint].webhookSubscription,
+  );
 }
 
 // 2020-07 onwards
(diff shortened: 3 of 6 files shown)


src/webhooks/registry.ts
```diff
@@ -74,6 +74,20 @@ function isSuccess(result: any, deliveryMethod: DeliveryMethod, webhookId?: stri
...
+            result.data.pubSubWebhookSubscriptionUpdate &&
+            result.data.pubSubWebhookSubscriptionUpdate.webhookSubscription,
+        );
+      } else {
+        return Boolean(
+          result.data &&
+            result.data.pubSubWebhookSubscriptionCreate &&
+            result.data.pubSubWebhookSubscriptionCreate.webhookSubscription,
+        );
+      }
```
james-langille, line 90: Minor: not sure if this makes it easier to read, but we could pull-out the only difference in these blocks to de-duplicate some code:
```suggestion
      endpoint = !!webhookId ? "pubSubWebhookSubscriptionUpdate" : "pubSubWebhookSubscriptionCreate"
      return Boolean(
        result.data &&
          result.data[endpoint] &&
          result.data[endpoint].webhookSubscription,
      );
```

Or, if we really want to help out, something like:
```javascript
endpoint = ""
switch (deliveryMethod) {
  case DeliveryMethod.HTTP:
    endpoint = "webhookSubscription";
    break;
  case DeliveryMethod.EventBridge:
    endpoint = "eventBridgeWebhookSubscription";
    break;
  case DeliveryMethod.PubSub:
    endpoint = "pubSubWebhookSubscription";
    break;
  default:
    return false;
}
endpoint += !!webhookId ? "Update" : "Create";
return Boolean(
  result.data &&
    result.data[endpoint] &&
    result.data[endpoint].webhookSubscription,
);
```
Thoughts on this?

src/webhooks/registry.ts
```diff
@@ -140,6 +166,8 @@ function buildQuery(
```
james-langille, line 166: I was going to say that we should just update the callers of this method to provide an object of the keys/values and validate it on our side rather than exposing the pubsub uri "format", but then I noticed that this function is exported and who knows how it's used.

In the future, we should probably split this function up into multiple, one per delivery method. But for now, this should be fine, and come to think of it, we don't necessarily need to have the `pubsub://` in front, just `<project>:<topic>` should work fine (something to consider).

tbrenev: Agreed.

Too much depends on that `path` argument right now. The most significant thing is that there's a local registry of webhooks, with each webhook having an associated handler function. I feel like it doesn't make sense to store EventBridge or PubSub webhooks in this registry (those webhooks are delivered to EventBridge/PubSub and handled there), so in the future, I'd suggest making an entirely separate function for registering non-HTTP webhooks, and deprecate this behavior in the current `register` function (so it will be used only for HTTP webhooks). However I think we need more context on how exactly this function is being used by partners.

src/webhooks/registry.ts
tbrenev: Good idea, thanks!

Went with the second option, that eliminates a lot of repetition.

james-langille (approved):

docs/usage/webhooks.md
```diff
@@ -74,6 +74,20 @@ app.get('/auth/callback', async (req, res) => {
 ```
 </details>
 
+### EventBridge and PubSub Webhooks
+
+You can also register webhooks for delivery to AWS EventBridge or Google Cloud
```
mikedasilva, line 79:
```suggestion
You can also register webhooks for delivery to Amazon EventBridge or Google Cloud
```

docs/usage/webhooks.md
```diff
@@ -74,6 +74,20 @@ app.get('/auth/callback', async (req, res) => {
...
+PubSub. In this case the `path` argument to
+`Shopify.Webhooks.Registry.register` needs to be of a specific form.
+
+For EventBridge, the `path` must be the [ARN of the partner event
+source](https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_EventSource.html).
+
+For PubSub, the `path` must be of the form
```
mikedasilva, line 86:
```suggestion
For Pub/Sub, the `path` must be of the form
```

docs/usage/webhooks.md
```diff
@@ -74,6 +74,20 @@ app.get('/auth/callback', async (req, res) => {
...
+`pubsub://[PROJECT-ID]:[PUB-SUB-TOPIC]`.  For example, if you created a topic
```
mikedasilva, line 87:
```suggestion
`pubsub://[PROJECT-ID]:[PUB-SUB-TOPIC-ID]`.  For example, if you created a topic
```

src/webhooks/registry.ts
```diff
@@ -45,48 +45,42 @@ interface RegistryInterface {
...
 
+function versionSupportsPubSub() {
+  return ShopifyUtilities.versionCompatible(ApiVersion.July21);
+}
+
 function validateDeliveryMethod(deliveryMethod: DeliveryMethod) {
   if (deliveryMethod === DeliveryMethod.EventBridge && !versionSupportsEndpointField()) {
     throw new ShopifyErrors.UnsupportedClientType(`EventBridge webhooks are not supported in API version "${Context.API_VERSION}".`);
+  } else if (deliveryMethod === DeliveryMethod.PubSub && !versionSupportsPubSub()) {
+    throw new ShopifyErrors.UnsupportedClientType(`PubSub webhooks are not supported in API version "${Context.API_VERSION}".`);
```
mikedasilva, line 83:
```suggestion
    throw new ShopifyErrors.UnsupportedClientType(`Pub/Sub webhooks are not supported in API version "${Context.API_VERSION}".`);
```

mikedasilva (approved):

src/webhooks/registry.ts
```diff
@@ -191,9 +199,11 @@ const WebhooksRegistry: RegistryInterface = {
...
       let endpointAddress = '';
       if ('endpoint' in node) {
-        endpointAddress = node.endpoint.__typename === 'WebhookHttpEndpoint'
-          ? node.endpoint.callbackUrl
-          : node.endpoint.arn;
+        if (node.endpoint.__typename === 'WebhookHttpEndpoint') {
+          endpointAddress = node.endpoint.callbackUrl;
+        } else if (node.endpoint.__typename === 'WebhookEventBridgeEndpoint') {
+          endpointAddress = node.endpoint.arn;
+        }
```
thecodepixi, line 206: Potentially silly question: is this just changing this from a ternary, or are we avoiding returning the incorrect thing for when we have a pub/sub hook with the more explicit `else if`?

thecodepixi (approved):
This looks really good! Just left a few nit-ish comments and one question for you.

src/base_types.ts
```diff
@@ -25,6 +25,7 @@ export enum ApiVersion {
   October20 = '2020-10',
   January21 = '2021-01',
   April21 = '2021-04',
+  July21 = '2021-07',
```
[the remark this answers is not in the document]
tbrenev: 👍 added!

src/webhooks/registry.ts
```diff
@@ -104,6 +98,12 @@ function buildCheckQuery(topic: string): string {
             ... on WebhookEventBridgeEndpoint {
               arn
             }
+            ${versionSupportsPubSub()
```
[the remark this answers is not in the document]
tbrenev: Yes it's a bit of a mess... unfortunately the GraphQL endpoint returns an error if we don't have this check. For older API versions which don't support EventBridge, it uses `legacyQuery` instead, which is why it doesn't error out, and why we don't need the check there. I think this could use a refactor (we should build a completely separate query for each delivery method?) but didn't want to risk touching too much in this PR.

src/webhooks/registry.ts
james-langille: Maybe another `switch` here would make that clearer?

PR #183, opened 2021-05-25
Fix incorrect example in webhook register docs

tbrenev:

Noticed that I made a small error in the docs in #181 - it's minor, but could be very confusing for someone reading the docs for the first time.

Just a tiny fix to the docs.

## Type of change

## Checklist

(don't think that we need a changelog entry for this, definitely don't need tests)


paulomarg (approved):
LGTM!

PR #196, opened 2021-06-11
Don't include extra params when calculating local hmac

t-kelly:

Encountered a situation where the returned query contained an extra key:value (`shopify: ['callback']`) which was then added to the local HMAC querystring and resulted in a false match and broken auth flow.

This change makes sure we are only passing the key:values we want when creating the local hmac.

## Type of change

## Checklist

### DIFF
--- a/src/auth/oauth/types.ts
+++ b/src/auth/oauth/types.ts
@@ -6,15 +6,6 @@ export interface AuthQuery {
   shop: string;
   host?: string;
   hmac?: string;
-  shopify?: string[];
-}
-
-export interface LocalHmacQuery {
-  code: string;
-  timestamp: string;
-  state: string;
-  shop: string;
-  host?: string;
 }
 
 export interface AccessTokenResponse {
--- a/src/utils/hmac-validator.ts
+++ b/src/utils/hmac-validator.ts
@@ -1,24 +1,24 @@
 import crypto from 'crypto';
 import querystring from 'querystring';
 
-import {AuthQuery, LocalHmacQuery} from '../auth/oauth/types';
+import {AuthQuery} from '../auth/oauth/types';
 import * as ShopifyErrors from '../error';
 import {Context} from '../context';
 
 import safeCompare from './safe-compare';
 
-export function stringifyQuery(query: LocalHmacQuery): string {
+export function stringifyQuery(query: AuthQuery): string {
   const orderedObj = Object.keys(query)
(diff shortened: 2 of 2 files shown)


src/utils/hmac-validator.ts
```diff
@@ -36,8 +36,8 @@ export default function validateHmac(query: AuthQuery): boolean {
       'Query does not contain an HMAC value.',
     );
   }
-  const {hmac, ...rest} = query;
-  const localHmac = generateLocalHmac(rest);
+  const {hmac, code, timestamp, state, shop} = query;
+  const localHmac = generateLocalHmac({code, timestamp, state, shop});
```
paulomarg, line 40: Should we include the `host` argument if it's here? I suspect this might fail if we're using App Bridge v2, but I could be wrong.

paulomarg (approved):
Amazing, thank you for this! I had a question around `host`, but once that's resolved I'm happy for this to be merged.

src/utils/hmac-validator.ts
t-kelly: Probably. I was working off of the types included in the lib because I forgot what values we were working with before. Will double check!

t-kelly: Done.

src/auth/oauth/types.ts
```diff
@@ -4,7 +4,17 @@ export interface AuthQuery {
   timestamp: string;
   state: string;
   shop: string;
+  host?: string;
   hmac?: string;
+  shopify?: string[];
+}
+
+export interface LocalHmacQuery {
```
paulomarg, line 12: Do we need the extra type here? We'll just ignore fields that we don't care about, so it wouldn't hurt to just use `AuthQuery` instead, would it?

src/utils/hmac-validator.ts
```diff
@@ -36,8 +36,8 @@ export default function validateHmac(query: AuthQuery): boolean {
       'Query does not contain an HMAC value.',
     );
   }
-  const {hmac, code, timestamp, state, shop} = query;
-  const localHmac = generateLocalHmac({code, timestamp, state, shop});
+  const {hmac, code, timestamp, state, shop, host} = query;
+  const localHmac = generateLocalHmac({code, timestamp, state, shop, ...host && {host}});
```
paulomarg, line 40: I also realized that we might want to do this in `generateLocalHmac` instead, to make sure we're always filtering out the keys we don't want.

PR #199, opened 2021-06-14
Add '/' route logic for Express app to docs

Paulinakhew:

- there is currently no '/' endpoint in the docs for Express

- add logic that checks if the user is logged in. If they're not logged in, redirects user to `login` with status code 302

## Type of change

## Checklist

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -187,6 +187,21 @@ While setting up `Context`, you'll be able to set [which version of the Admin AP
     console.log('your app is now listening on port 3000');
   });
   ```
+  
+  You will also need to set up a basic router to be able to process requests:
+
+  ```ts
+  
+  app.get('/', async (req, res) => {
+    const currentSession = await Shopify.Utils.loadCurrentSession(req, res);
+    if (!currentSession) {
+      // not logged in, redirect to login
+      return res.redirect(302, '/login');
+    } else {
+      // do something amazing with your application!
+    }
+  })
+  ```
 </details>
 
 ## Running your app


docs/getting_started.md
```diff
@@ -187,6 +187,21 @@ While setting up `Context`, you'll be able to set [which version of the Admin AP
     console.log('your app is now listening on port 3000');
   });
   ```
+  
+  You will also need to set up a basic router to be able to process requests:
+
+  ```ts
+  
+  app.get('/', async (req, res) => {
+    const currentSession = await Shopify.Utils.loadCurrentSession(req, res);
```
paulomarg, line 196: We shouldn't be calling `loadCurrentSession` here, that won't work for embedded apps. Maybe we should fix that here? An extremely simple way of doing this with app bridge v2 would be to check for `host` in the URL, since this request is essentially unauthenticated - i.e. if it loads inside of the admin, we can let it through and require permissions when we need them.

docs/getting_started.md
```diff
@@ -187,6 +187,21 @@ While setting up `Context`, you'll be able to set [which version of the Admin AP
...
+    if (!currentSession) {
+      // not logged in, redirect to login
+      return res.redirect(302, '/login');
+    } else {
+      // do something amazing with your application!
+    }
+  })
+  ```
```
paulomarg, line 204: Let's add some indication that this needs to happen before the listen call (I think).

paulomarg (commented):
Looking good!

bishpls:
I have some feedback on this: as-is, the documentation is fundamentally broken for standalone apps using offline sessions for a couple of reasons, and it'd be a pretty huge quality of life improvement for the correct way of handling it to be included in the official guides!

First issue: loadCurrentSession in this scenario is flat-out broken, due to the bug mentioned in this PR https://github.com/Shopify/shopify-node-api/pull/272. As such, we'll use loadOfflineSession.

Second issue: loadOfflineSession doesn't validate the shop parameter to confirm the route request is actually coming from Shopify; as such, we'll need to validate the hmac before proceeding with the route. 

Third issue: the built-in Shopify.Utils.validateHmac function doesn't work in this context (as the query string keys are different from those defined in the AuthQuery constructor -- we have hmac, host, **session**, shop, and timestamp here as opposed to AuthQuery's **code**, timestamp, **state**, shop, host, and hmac); as such, a custom validation function is required. 

Fourth issue: Due to the fact that beginAuth generates an un-auth'd session (I.E. no accessToken or scopes) the moment the install request is generated but BEFORE oAuth has actually been confirmed, the existence of a session actually does not at all determine whether or not a user should be directed to the /login (I.E. generate oAuth) route. See: https://github.com/Shopify/shopify-node-api/issues/255. ​I don't believe it would be proper to just discard the un-auth'd temp session generated by beginAuth, as validateAuthCallback requires a session to proceed.

I believe that the way to handle checking for whether a user should be directed to /login (I.E. beginAuth, followed by validateAuthCallback) is whether the loaded session has an accessToken, not just that the session exists. (Since it's entirely possible for a user to abandon the install flow, leaving an un-authorized session in perpetual limbo). Most of the example code uses the ACTIVE_SHOPIFY_SHOPS local object (with the assumption that this is loaded from the database on server start) to approximate 'shop has finished oAuth,' but it seems like accessToken is a more robust measure of this? (A more complete solution ought to also check that scopes match, I believe)


As best I can tell, setting up a minimally-viable default-route endpoint (and redirecting to a login path as necessary) using offline sessions should look something like the following, which I've tested on my own app for both post-install oAuth routing and 'Vendor clicks on the app link from admin/apps' routing:

`  api.get('/', async(req, res) => {
   ​const shop = req.query.shop;

   ​const {hmac, ...cleanQueryKeys} = req.query;
   ​const cleanQueryString = new URLSearchParams(cleanQueryKeys).toString();
              ​
   ​const generatedHash = crypto.createHmac('SHA256', Shopify.Context.API_SECRET_KEY).update(cleanQueryString).digest('hex');
                                                                                                          ​
   ​const safeCompareResult = Shopify.Utils.safeCompare(generatedHash, hmac);
                               ​
   ​if (!safeCompareResult) {   
     ​console.log('Invalid hmac, aborting');
     ​return res.status(401).json({ succeeded: false, message: 'Not Authorized' }).send();
   ​}  

   ​const authSession = await Shopify.Utils.loadOfflineSession(shop);
   ​if (!(authSession && authSession.accessToken)) {
     ​console.log('Shop auth is not yet authorized, redirecting to login');
     ​res.redirect(`login?shop=${shop}`);
   ​} else {
     ​console.log('Shop is authorized, redirecting to standalone app');
     ​return res.redirect(`https://example.app.com?redirect=shopify&hmac=${hmac}&shop=${shop}`);
   ​}
 ​});

PR #204, opened 2021-06-25
Update "Make a REST API call" documentation

hta218:
Help other devs not wasting time while figuring out how to add `query` params


hta218:
Already signed the CLA

image

But why does the `probot` still ask me to sign it?

image

PR #205, opened 2021-06-25
Add addHandler and getHandler functions

Paulinakhew:

- Refactored the `register` function

- added a console.log that lets the 3p developer know that the method is deprecated for `loadHandler` with handler
- extract a `loadHandler` method out of `register` that just adds the handler to the registry, and make the handler parameter optional in `register` to not break things

## Type of change

## Checklist

### DIFF
--- a/src/webhooks/registry.ts
+++ b/src/webhooks/registry.ts
@@ -22,6 +22,20 @@ interface RegistryInterface {
   webhookRegistry: WebhookRegistryEntry[];
 
   /**
+   * Sets the handler for the given topic. If a handler was previously set for the same topic, it will be overridden.
+   *
+   * @param options Paramters to add a handler which are topic and webHookHandler
+   */
+  addHandler(options: WebhookRegistryEntry): Promise<void>;
+
+  /**
+   * Fetches the handler for the given topic. Returns null if no handler was registered.
+   *
+   * @param topic The topic to check
+   */
+  getHandler(topic: string): Promise<WebhookRegistryEntry | null>;
+
+  /**
    * Registers a Webhook Handler function for a given topic.
    *
    * @param options Parameters to register a handler, including topic, listening address, handler function
@@ -177,6 +191,16 @@ function buildQuery(
 const WebhooksRegistry: RegistryInterface = {
   webhookRegistry: [],
 
+  async addHandler({path, topic, webhookHandler}): Promise<void> {
+    WebhooksRegistry.webhookRegistry = WebhooksRegistry.webhookRegistry.filter((item) => item.topic !== topic);
+    WebhooksRegistry.webhookRegistry.push({path, topic, webhookHandler});
+  },
+
+  async getHandler(topic) {
+    const webhookEntry = WebhooksRegistry.webhookRegistry.find((entry) => entry.topic === topic);
+    return (webhookEntry) ? webhookEntry : null;
+  },
+
   async register({
     path,
     topic,
@@ -185,6 +209,9 @@ const WebhooksRegistry: RegistryInterface = {
     deliveryMethod = DeliveryMethod.Http,
     webhookHandler,
   }: RegisterOptions): Promise<RegisterReturn> {
+    if (webhookHandler) console.log(
+      'This method of passing in the webhook is now deprecated. Please use the addHandler function.'
+    );
     validateDeliveryMethod(deliveryMethod);
     const client = new GraphqlClient(shop, accessToken);
     const address = deliveryMethod === DeliveryMethod.Http
@@ -225,12 +252,6 @@ const WebhooksRegistry: RegistryInterface = {
     } else {
       success = true;
       body = {};
-    }
-
-    if (success) {
-      // Remove this topic from the registry if it is already there
-      WebhooksRegistry.webhookRegistry = WebhooksRegistry.webhookRegistry.filter((item) => item.topic !== topic);
-      WebhooksRegistry.webhookRegistry.push({path, topic, webhookHandler});
     }
 
     return {success, result: body};
--- a/src/webhooks/types.ts
+++ b/src/webhooks/types.ts
@@ -16,7 +16,7 @@ export interface RegisterOptions {
   path: string;
   shop: string;
   accessToken: string;
-  webhookHandler: WebhookHandlerFunction;
+  webhookHandler?: WebhookHandlerFunction;
   deliveryMethod?: DeliveryMethod;
 }
 


src/webhooks/registry.ts
```diff
@@ -175,15 +189,23 @@ function buildQuery(
 }
 
 const WebhooksRegistry: RegistryInterface = {
-  webhookRegistry: [],
+  webhookRegistry: {},
+
+  async addHandler({path, topic, webhookHandler}): Promise<void> {
```
paulomarg, line 194: I think we need to type these params as well:

```suggestion
  async addHandler({path, topic, webhookHandler}: WebhookRegistryEntry): Promise<void> {
```

src/webhooks/registry.ts
```diff
@@ -297,7 +313,7 @@ const WebhooksRegistry: RegistryInterface = {
 
         if (ShopifyUtilities.safeCompare(generatedHash, hmac as string)) {
           const graphqlTopic = (topic as string).toUpperCase().replace(/\//g, '_');
-          const webhookEntry = WebhooksRegistry.webhookRegistry.find((entry) => entry.topic === graphqlTopic);
+          const webhookEntry = WebhooksRegistry.webhookRegistry[graphqlTopic];
```
paulomarg, line 316: Should this be a call to `getHandler` instead?

src/webhooks/test/registry.test.ts
```diff
@@ -397,10 +385,10 @@ describe('ShopifyWebhooks.Registry.register', () => {
       topic: 'PRODUCTS_CREATE',
       accessToken: 'some token',
       shop: 'shop1.myshopify.io',
-      webhookHandler: genericWebhookHandler,
     };
+    await ShopifyWebhooks.Registry.addHandler({path: '/webhooks', topic: 'PRODUCTS_CREATE', webhookHandler: genericWebhookHandler});
     await ShopifyWebhooks.Registry.register(webhook);
-    expect(ShopifyWebhooks.Registry.webhookRegistry).toHaveLength(1);
+    expect(Object.keys(ShopifyWebhooks.Registry.webhookRegistry).length).toBe(1);
```
paulomarg, line 391: Totally a nit, but you could do this:
```suggestion
    expect(Object.keys(ShopifyWebhooks.Registry.webhookRegistry)).toHaveLength(1);
```

src/webhooks/test/registry.test.ts
```diff
@@ -443,15 +428,15 @@ describe('ShopifyWebhooks.Registry.process', () => {
   });
 
   afterEach(async () => {
-    ShopifyWebhooks.Registry.webhookRegistry = [];
+    ShopifyWebhooks.Registry.webhookRegistry = {};
   });
 
   it('handles the request when topic is already registered', async () => {
-    ShopifyWebhooks.Registry.webhookRegistry.push({
+    ShopifyWebhooks.Registry.webhookRegistry.PRODUCTS = {
```
paulomarg, line 435: We should replace all of these calls that are directly setting values to `webhookRegistry` to call `addHandler` instead to test the method.

src/webhooks/test/registry.test.ts
```diff
@@ -656,6 +641,54 @@ describe('ShopifyWebhooks.Registry.isWebhookPath', () => {
   });
 });
 
+describe('ShopifyWebhooks.Registry.addHandler', () => {
+  beforeEach(async () => {
+    ShopifyWebhooks.Registry.webhookRegistry = {};
+  });
+
+  it('adds two handlers to the webhook registry', async () => {
+    await ShopifyWebhooks.Registry.addHandler({path: '/webhooks', topic: 'PRODUCTS_CREATE', webhookHandler: genericWebhookHandler});
```
paulomarg, line 650: Nit: is the linter OK with such long lines? It probably shouldn't be 😅  - we might want to fix that as a separate PR.

src/webhooks/registry.ts
```diff
@@ -19,12 +19,26 @@ import {
...
+   * @param options Paramters to add a handler which are path, topic and webHookHandler
+   */
+  addHandler(options: WebhookRegistryEntry): Promise<void>;
+
+  /**
+   * Fetches the handler for the given topic. Returns null if no handler was registered.
+   *
+   * @param topic The topic to check
+   */
+  getHandler(topic: string): Promise<WebhookRegistryEntry | null>;
```
paulomarg, line 36: I'm pretty sure these methods don't need to return Promises, since they're just doing dictionary operations which don't need to be deferred.

src/webhooks/registry.ts
```diff
@@ -175,15 +189,23 @@ function buildQuery(
...
 const WebhooksRegistry: RegistryInterface = {
-  webhookRegistry: [],
+  webhookRegistry: {},
+
+  async addHandler({path, topic, webhookHandler}): Promise<void> {
+    delete WebhooksRegistry.webhookRegistry[topic];
+    WebhooksRegistry.webhookRegistry[topic] = {path, topic, webhookHandler};
+  },
+
+  async getHandler(topic) {
```
paulomarg, line 199: We're lacking the typing info here :)

paulomarg (commented):
Looking great! Had a couple of comments but the methods themselves are pretty much good to go!

src/webhooks/registry.ts
Paulinakhew: ohhh this would make my life much easier

src/webhooks/test/registry.test.ts
Paulinakhew: Yeah I didn't know if I should do that because if the `addHandler` function breaks it would break the other tests. This makes more sense though because if the `addHandler` function breaks everything else would break anyways

src/webhooks/test/registry.test.ts
Paulinakhew: this is so much nicer

src/webhooks/registry.ts
Paulinakhew: yess

src/webhooks/registry.ts
```diff
@@ -202,15 +216,22 @@ function buildQuery(
 }
 
 const WebhooksRegistry: RegistryInterface = {
-  webhookRegistry: [],
+  webhookRegistry: {},
+
+  addHandler({path, topic, webhookHandler}: WebhookRegistryEntry): void {
+    WebhooksRegistry.webhookRegistry[topic] = {path, topic, webhookHandler};
+  },
```
mkevinosullivan, line 223: Since `topic` is now the index into the hash, it's no longer needed in the `WebhookRegistryEntry` type itself, i.e., in the `webhooks/types.ts` file, the interface can be changed to
```
export interface WebhookRegistryEntry {
  path: string;
  webhookHandler: WebhookHandlerFunction;
}
```

I've had a look through the rest of the code in `registry.ts` and `<entry>.topic` isn't used anywhere, as far as I can see.

Paulinakhew: I don't think I can remove it because in `addHandler` I need the `topic` to use as a key in the `WebhooksRegistry.webhookRegisry` object

```js
  addHandler({path, topic, webhookHandler}: WebhookRegistryEntry): void {
    WebhooksRegistry.webhookRegistry[topic] = {path, topic, webhookHandler};
  },
```

what I can do instead is remove the topic from the values since it's redundant

pveen2:
> ```ts
> Shopify.Webhooks.Registry.reloadHandlers();
> 
>   try {
>     await Shopify.Webhooks.Registry.process(req, res);
>     console.log(`Webhook processed, returned status code 200`);
>   } catch (error) {
>     console.log(`Failed to process webhook: ${error}`);
>   }
> ```

+1 Experiencing exactly the same in a NEXT.js app, did you solve it @cmnstmntmn

PR #209, opened 2021-07-02
Add support for custom location protocol

paul-phan:

I'm on local development and don't want to use `ngrok` to publish my NodeJS server because it might slow sometimes.
Add this protocol support to allow us to use `localhost:3000` without the need of faking an SSL certificate.

Added `PROTOCOL` to the `Shopify.Context`, by default it always is `https:`

## Type of change

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,8 @@ coverage/
 yarn-error.log
 tsdocs/
 tsconfig.tsbuildinfo
+/.idea/inspectionProfiles/Project_Default.xml
+/.idea/.gitignore
+/.idea/modules.xml
+/.idea/shopify-node-api.iml
+/.idea/vcs.xml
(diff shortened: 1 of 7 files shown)


.gitignore
```diff
@@ -6,3 +6,8 @@ coverage/
 yarn-error.log
 tsdocs/
 tsconfig.tsbuildinfo
+/.idea/inspectionProfiles/Project_Default.xml
+/.idea/.gitignore
+/.idea/modules.xml
+/.idea/shopify-node-api.iml
+/.idea/vcs.xml
```
Paulinakhew, line 13: Hi sorry can you remove these? These seem to be specific to your local dev environment

src/base_types.ts
```diff
@@ -6,6 +6,7 @@ export interface ContextParams {
   API_SECRET_KEY: string;
   SCOPES: string[] | AuthScopes;
   HOST_NAME: string;
+  PROTOCOL?: 'http:' | 'https:';
```
Paulinakhew, line 9:
```suggestion
  PROTOCOL?: string;
```
You should replace the strings with the string type

Paulinakhew (changes requested):
Looks good! There are just a couple of things that need to be modified

PR #210, opened 2021-07-06
Add linting rule for single quotes and max line length of 120

Paulinakhew:

- linter is not enforcing a max line length of `120`
- [JS shopify style guide here](https://github.com/Shopify/javascript)

- add max line length rule
- add single quotes rule
- reformat code

## Type of change

## Checklist

### DIFF
--- a/src/clients/http_client/http_client.ts
+++ b/src/clients/http_client/http_client.ts
@@ -211,18 +211,18 @@ class HttpClient {
             case response.status === StatusCode.TooManyRequests: {
               const retryAfter = response.headers.get('Retry-After');
               throw new ShopifyErrors.HttpThrottlingError(
-              `Shopify is throttling requests${errorMessage}`,
-              retryAfter ? parseFloat(retryAfter) : undefined,
-            );
+                `Shopify is throttling requests${errorMessage}`,
+                retryAfter ? parseFloat(retryAfter) : undefined,
+              );
             }
             case response.status >= StatusCode.InternalServerError:
               throw new ShopifyErrors.HttpInternalError(`Shopify internal error${errorMessage}`);
             default:
               throw new ShopifyErrors.HttpResponseError(
-              `Received an error response (${response.status} ${response.statusText}) from Shopify${errorMessage}`,
-              response.status,
-              response.statusText,
-            );
+                `Received an error response (${response.status} ${response.statusText}) from Shopify${errorMessage}`,
+                response.status,
+                response.statusText,
+              );
           }
         }
       })


src/auth/oauth/test/oauth.test.ts
```diff
@@ -266,7 +267,8 @@ describe('validateAuthCallback', () => {
     expect(session?.accessToken).toBe(successResponse.access_token);
   });
 
-  test('requests access token for valid callbacks with online access and updates session with expiration and onlineAccessInfo', async () => {
+  test('requests access token for valid callbacks with online access and updates session' +
+    ' with expiration and onlineAccessInfo', async () => {
```
paulomarg, line 271: I don't love breaking the string here, but we can always rewrite it something more sensibly long if we ever refactor this test.

src/clients/http_client/http_client.ts
```diff
@@ -211,18 +211,18 @@ class HttpClient {
             case response.status === StatusCode.TooManyRequests: {
               const retryAfter = response.headers.get('Retry-After');
               throw new ShopifyErrors.HttpThrottlingError(
-                `Shopify is throttling requests${errorMessage}`,
-                retryAfter ? parseFloat(retryAfter) : undefined,
-              );
+              `Shopify is throttling requests${errorMessage}`,
+              retryAfter ? parseFloat(retryAfter) : undefined,
+            );
```
paulomarg, line 216: This indentation here feels very counter-intuitive to me. Is it part of the new rules?

src/test/context.test.ts
```diff
@@ -36,7 +36,7 @@ describe('Context object', () => {
     expect(Context.HOST_NAME).toEqual(validParams.HOST_NAME);
   });
 
-  it("can't initialize with empty values", () => {
+  it('can\'t initialize with empty values', () => {
```
paulomarg, line 39: I thought double quotes were allowed in this scenario? Did this fail with the rule?

paulomarg (commented):
This all makes sense to me, just had a couple of general questions.

src/clients/http_client/http_client.ts
Paulinakhew: ohh I was trying out an indentation rule and used the `--fix` flag. I'll revert these changes because they were better before

JaKXz:
@Paulinakhew instead of manually trying to figure out the formatting, I would strongly suggest adding 
```
yarn add -D prettier @shopify/prettier-config
```
and then applying the formatting to all files with:

```
npx prettier --write '**/*'
```

and then modifying the [eslint config](https://github.com/Shopify/web-configs/tree/main/packages/eslint-plugin) and fixing linting as a result. It might also be worth adding a `dev.yml` with a pre-commit hook configured.

I love chores like this, thank you for taking it up.

PR #213, opened 2021-07-07
Clean up code using @shopify/prettier-config

Paulinakhew:

- inspired by [this comment](https://github.com/Shopify/shopify-node-api/pull/210#issuecomment-875700298) on my add linting PR
- this PR will be followed up with pre-commit and more linting rules (for example, the max line length still isn't enforced through `yarn lint`)

- ran `npx prettier` on all files

## Type of change

## Checklist


JaKXz (approved):
Nice! 

I think with this and the follow up to enforce prettier in CI you won't have to worry about formatting rules like line length :)

precha1989 (commented):
good

PR #217, opened 2021-07-13
Return and delete session in validateAuthCallback

Paulinakhew:

- Consider whether it is possible to never fall back to loading the current session from a cookie on embedded apps (and make sure we expire the OAuth cookie right away).

- returns the current session at the end of `validateAuthCallback`
- immediately destroy the cookie session for embedded apps to help prevent build-up

## Type of change

## Checklist

## Note to self
```
    // Begin auth
    // Check that we DO have a cookie (that expires 1m in the future with grace period)
    // Set up the test callback query
    // Call validateAuthCallback
    // Check that the cookie has the right id
    // Check that the cookie has the right expiration date
    // expect(cookies.expires).toBe("the right value");
```

### DIFF
--- a/src/auth/oauth/test/oauth.test.ts
+++ b/src/auth/oauth/test/oauth.test.ts
@@ -470,6 +470,7 @@ describe('validateAuthCallback', () => {
     expect(returnedSession.id).toEqual(cookies.id);
 
     expect(returnedSession?.expires?.getTime() as number).toBeWithinSecondsOf(new Date(Date.now() + successResponse.expires_in * 1000).getTime(), 1);
+    expect(cookies?.expires?.getTime() as number).toBeWithinSecondsOf(returnedSession?.expires?.getTime() as number, 1);
 
     const cookieSession = await Context.SESSION_STORAGE.loadSession(cookies.id);
     expect(cookieSession).not.toBeUndefined();
@@ -517,6 +518,7 @@ describe('validateAuthCallback', () => {
     const cookieSession = await Context.SESSION_STORAGE.loadSession(cookies.id);
     expect(cookieSession).not.toBeUndefined();
     expect(cookies?.expires?.getTime() as number).toBeWithinSecondsOf(new Date().getTime(), 1);
+    expect(returnedSession?.expires?.getTime()).toBeUndefined();
   });
 
   test('properly updates the Oauth cookie for offline, non-embedded apps', async () => {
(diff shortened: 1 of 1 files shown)


src/auth/oauth/oauth.ts
```diff
@@ -182,9 +184,12 @@ const ShopifyOAuth = {
       const jwtSession = Session.cloneSession(currentSession, jwtSessionId);
       await Context.SESSION_STORAGE.storeSession(jwtSession);
 
-      // Make sure the current OAuth session expires along with the cookie
-      oauthSessionExpiration = new Date(Date.now() + 30000);
-      currentSession.expires = oauthSessionExpiration;
+      const sessionDeleted = Context.SESSION_STORAGE.deleteSession(currentSession.id);
```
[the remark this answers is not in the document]
paulomarg: Missed this before, but we need to `await` this call.

```suggestion
      const sessionDeleted = await Context.SESSION_STORAGE.deleteSession(currentSession.id);
```

src/auth/oauth/test/oauth.test.ts
```diff
@@ -370,21 +370,11 @@ describe('validateAuthCallback', () => {
     testCallbackQuery.hmac = expectedHmac;
 
     fetchMock.mockResponse(JSON.stringify(successResponse));
-    await ShopifyOAuth.validateAuthCallback(req, res, testCallbackQuery);
+    const returnedCookie = await ShopifyOAuth.validateAuthCallback(req, res, testCallbackQuery);
```
paulomarg, line 373: This is great, but I feel like `returnedCookie` doesn't really represent this.
```suggestion
    const currentSession = await ShopifyOAuth.validateAuthCallback(req, res, testCallbackQuery);
```

src/auth/oauth/test/oauth.test.ts
```diff
@@ -370,21 +370,11 @@ describe('validateAuthCallback', () => {
...
-
-    if (cookieSession?.expires) {
-      const actualCookieExpiration: number =
-        cookieSession.expires.getTime() / 1000;
-      // 1-second grace period
-      expect(
-        Math.abs(expectedCookieExpiration - actualCookieExpiration),
-      ).toBeLessThan(1);
-    }
+    expect(cookieSession).toEqual(returnedCookie);
```
paulomarg, line 377: This session should actually be deleted now - `loadSession` will return it but the expiration should be in the past.

src/auth/oauth/oauth.ts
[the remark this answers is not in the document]
Paulinakhew: big brain 🤯

Paulinakhew:
Hi @paulomarg I still need to write tests but I'm re-requesting your review

src/auth/oauth/test/oauth.test.ts
```diff
@@ -25,6 +24,31 @@ beforeEach(() => {
   (Cookies as any).mockClear();
 });
 
+declare global {
+  interface Matchers<R> {
+    toBeWithinSecondsOf(compareDate: number, seconds: number): R;
+  }
+}
+
+expect.extend({
```
paulomarg, line 35: I'd suggest moving this to the test helper since it's pretty useful throughout all tests!

PR #221, opened 2021-07-26
Update webhooks.md - method uses incorrect prop variables

mr-moto:
Register webhooks section (Express) loadCurrentSession method uses `request` and `response` props instead of `req` and `res` as returned in the callback

Fixes a typo in the docs

## Type of change

- [x] Doc Update


mr-moto:
cla signed

thecodepixi (approved):
Good catch! Thanks!

PR #226, opened 2021-07-31
Move Webhooks registry to Context

arsnl:

The goal of this PR is to abstract the `Webhooks registry` from the `Webhooks.register` and `Webhooks.process` methods since they have different responsibilities.

- The `Webhooks registry` responsibility is to describe the different topics the application need to listen and the way those topics will be handled.
- The `Webhooks.register` method responsibility is to inform Shopify about those topics the application need to listen and where to send the payload, based on the `Webhooks registry` informations.
- The `Webhooks.process` method responsibility is to handle the Webhooks with the `HTTP deliveryMethod`.

Currently, the `Webhooks registry` only exist in memory (in `Shopify.Webhooks.Registry.webhookRegistry`) after a complete OAuth and is populated via the `Webhooks.register` method.

The issue with this approach is that `Shopify.Webhooks.Registry.webhookRegistry` is only temporary. Which is not ideal with modern JAMStack/Serverless/Stateless applications. Even on a classic monolithic server, after a restart the registry is lost.

For example, on Serverless, after an install of the application on a store the endpoint `/auth/callback` is called, the `Webhooks registry` method is executed, the Webhooks are registered on Shopify and the registry is populated. However, if the user uninstall the application, the endpoint `/webhooks` is now called. Since every calls are stateless, the memory is now empty. The `Webhooks process` method is executed, looking for the topic `APP_UNINSTALLED` in the now empty `Shopify.Webhooks.Registry.webhookRegistry` an since the registry is empty the server return an error and don't execute the app uninstallation processes.

#### `WEBHOOKS_REGISTRY`

Add the `WEBHOOKS_REGISTRY` property to the Context. This property is optional and default to `{}` if missing.

```ts
interface WebhookRegistry {
  [topic: string]: {
    path: string;
    webhookHandler: WebhookHandlerFunction;
    deliveryMethod?: DeliveryMethod;
  };
}
```

Also, the memory registry (`Shopify.Webhooks.Registry.webhookRegistry`) was removed since `WEBHOOKS_REGISTRY` replace it in the Context.

#### `Webhooks.register`

The properties `topic`, `path`, `webhookHandler` and `deliveryMethod` was removed from the `Webhooks.register` options to provide since they are now directly inside the `WEBHOOKS_REGISTRY`. The only (and required) options to provide to `Webhooks.register` are the `shop` and `accessToken`.

When the `Webhooks.register` method is executed, all the Webhooks present inside the `WEBHOOKS_REGISTRY` are registered to Shopify without having to provide them manually to the register method. That mean no more need to call `Webhooks.register` for every Webhook we need to register.

Also, return format of `Webhooks.register` is passed from `Promise<{success, result}>` to `Promise<{[topic]: {success, result}}>`

```ts
// Initialize the context higher in the process
const WEBHOOKS_REGISTRY = {
  APP_UNINSTALLED: {
    path: '/webhooks',
    webhookHandler: async (topic, shop, body) => {
      console.log('APP_UNINSTALLED handler was executed');
    }
  },
  PRODUCTS_CREATE: {
    path: '/webhooks',
    webhookHandler: async (topic, shop, body) => {
      console.log('PRODUCTS_CREATE handler was executed');
    }
  }
}

Shopify.Context.initialize({
  API_KEY: process.env.SHOPIFY_API_KEY,
  API_SECRET_KEY: process.env.SHOPIFY_API_SECRET,
  SCOPES: process.env.SCOPES.split(","),
  HOST_NAME: process.env.HOST.replace(/https:\/\//, ""),
  IS_EMBEDDED_APP: true,
  API_VERSION: ApiVersion.July21,
  WEBHOOKS_REGISTRY
});

// ...

// Register Webhooks after OAuth completes
app.get('/auth/callback', async (req, res) => {
  try {
    await Shopify.Auth.validateAuthCallback(req, res, req.query); 

    const { accessToken, shop } = await Shopify.Utils.loadCurrentSession(req, res);

    const registerReturn = await Shopify.Webhooks.Registry.register({ accessToken, shop });
    console.log(registerReturn); //=> { APP_UNINSTALLED: { success: true, result: ... }, PRODUCTS_CREATE: { success: true, result: ... } }
  } catch (error) {
    console.error(error); // in practice these should be handled more gracefully
  }
  return res.redirect('/'); // wherever you want your user to end up after OAuth completes
});
```

#### `Webhooks.process` and `Webhooks.isWebhookPath`

Those methods usage did not change for the user, but internally they are now looking inside the key-value `WEBHOOKS_REGISTRY` and not inside the array `Shopify.Webhooks.Registry.webhookRegistry` to grab the handler.

The interesting thing is now the users wont have to do hack's (see: https://github.com/Shopify/shopify-node-api/issues/157#issuecomment-841172584) to make sure `Webhooks.process` and `Webhooks.isWebhookPath` methods doesn't fail.

#### `Context.initialize`

Add the `WEBHOOKS_REGISTRY` property. I've also enforce the default values, validation of other properties.

## Type of change

## Checklist


chriszirkel:
I just stumbled upon this issue myself. I didn't expect that the webhook registry will be only kept in memory and will not be populated automatically within Context initialize.  This is clearly an architectural error and I'm hoping the team will accept your PR.

p0fi:
For gods sake, someone please accept this PR!

paulomarg:
Hey folks! First of all, let me start off by apologizing for the long wait on this one.

We've been working on a handful of improvements to this library that would be breaking changes so we can release a v2 and deploy all of those breaking changes at once to minimize the disruption to apps using it. One of those changes is https://github.com/Shopify/shopify-node-api/pull/205, which @arsnl has provided some valuable insight on (thanks for that!).

We did take that feedback into consideration, and we changed a few things in that PR which will hopefully make life easier for your apps, mainly that we wanted to provide an `addHandlers` method which would allow an app to do something like this when setting up your server / vercel apps:

```js
Shopify.Context.initialize({ ... });
Shopify.Webhooks.addHandlers({ topic1: handler1, topic2: handler2 }); // Not necessarily in the same file as above
```

Our goal with that is to make it easy for an app to register all of the webhooks it'll need at startup and not have to worry about the handlers later. After OAuth you can just go over the topics you need to process and call `register` - we're also considering adding a `registerAll` method to the webhooks component that will just loop over the handlers that were added and make sure they're all set up if they'd already been `add`ed into the registry.

While it doesn't fully move the handlers into `Context`, it will have the same effect of making webhook registration a 1-call deal while keeping the handlers in a specific object within the library, which helps prevent bloating the `Context`. You can also save the handlers in a separate file and call `addHandlers` for them at any point in the app so they don't have to be in memory.

We're working out the last kinks in our PR before we can merge it and release the new version, but we'd definitely appreciate your feedback on whether this approach would work for your apps.

Thanks again for your patience!

cmnstmntmn:
@paulomarg when will this be merged?

arsnl:
@danielbixel great question! 

I didn't updated this PR since @paulomarg said the core team was working on a similar solution for the release of v2. I didn't use the shopify-node-api in a while since it was not really interesting for modern stack so I don't know the evolution of this project.

@paulomarg do you want me to update this PR or did you find a solution to support a stateless registry in the v2?

paulomarg:
As part of the v2 updates, we separated the webhook registration into 2 steps:

1. Adding handlers for HTTP webhooks to the registry. This essentially amounts to what this PR is doing - though with a slightly different implementation since the handlers still go in the registry rather than the Context. This can happen at any time before the `register` call, for example when the `Context` is set up.
2. Actually triggering the call to register the webhooks with Shopify via `register` or `registerAll`. This is the only action that needs to happen after OAuth, so you no longer need to define your HTTP handlers as part of the auth process.

Given that the end result of both changes is quite similar, I think this PR could be closed. We're happy to improve on our current solution if it is still has problems.

Please find more details in the [webhooks docs](https://github.com/Shopify/shopify-node-api/blob/main/docs/usage/webhooks.md).

PR #258, opened 2021-11-10
Fix transient json parse error

### LINKED ISSUE #257: Transient shopify error (for html response instead of json) is thrown as HttpRequestError instead of HttpRetriableError
# Issue summary

Sometimes query returns non json response (probably html but only parsing error is logged). 
This error is wrongly interpreted as request error and isn't retried.

## Expected behavior

`HttpRetriableError` should be thrown instead of `HttpRequestError`

## Actual behavior

When we get non json response [HttpRequestError](https://github.com/Shopify/shopify-node-api/blob/975b87985c57cc723488a9f712de39a28ce2c967/src/clients/http_client/http_client.ts#L264-L266) which is caused by [response.json()](https://github.com/Shopify/shopify-node-api/blob/975b87985c57cc723488a9f712de39a28ce2c967/src/clients/http_client/http_client.ts#L180) parsing `HttpRequestError` is thrown.

<details>
Failed to make Shopify HTTP request: FetchError: invalid json response body at https://XXX.myshopify.com/admin/api/2021-07/products.json?limit=250&page_info=XXX reason: Unexpected token < in JSON at position 0
</details>

It's hard to provide reproduce steps for transient error but I'll provide PR with this situation handled in test case.

wszaranski:
For all shopify requests we expect to receive json response. When non
json response is received `HttpRequestError` was thrown instead of
`HttpRetriableError`.

FIXES #257
## Type of change

## Checklist

### DIFF
--- a/src/clients/http_client/__tests__/http_client.test.ts
+++ b/src/clients/http_client/__tests__/http_client.test.ts
@@ -298,6 +298,7 @@ describe('HTTP client', () => {
         {status: 500, statusText, headers: {'x-request-id': requestId}},
       ],
       ['<not a json response>', {status: 500, statusText: 'Did not work'}],
+      ['<not a json response>', {status: 200}],
     );
 
     await testErrorResponse(403, ShopifyErrors.HttpResponseError, true);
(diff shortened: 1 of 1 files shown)


src/clients/http_client/http_client.ts
```diff
@@ -260,6 +260,8 @@ class HttpClient {
       .catch((error) => {
         if (error instanceof ShopifyErrors.ShopifyError) {
           throw error;
+        } else if (typeof error === 'object' && error.name === 'FetchError' && error.type === 'invalid-json') {
```
codeallthethingz, line 263: I might extract this complicated if statement into a function 

`isRetriableError(error)`

JaKXz:
Hey @wszaranski thanks for the PR, but we've decided to go with the solution here in https://github.com/Shopify/shopify-node-api/pull/398. I'm going to close this for now.

PR #269, opened 2021-12-12
There is a typo as it should be res.send()

AmJustSam:
There is a type Inside `app.get()`, we are using `response.end()` when it should be `res.end()` based on the parameters we are using.


paulomarg (approved):
Great catch, thanks!

AmJustSam:
I'm getting this error while trying to sign it "Submission rate limit exceeded. Please try again later".

mllemango:
I've rerun the CLA check, could you try again? If its still not working could you push an empty commit to re-trigger the check?

AmJustSam:
Done, signed.

PR #272, opened 2021-12-14
[Bug] Remove cookie check when validating offline sessions

fjij:

Offline sessions don't need or require cookies in order to be used, yet `validateAuthCallback` always checks for cookies.

This makes it difficult to use offline sessions in a scenario where cookies cannot be used, despite it being possible.

Add a `isOnline` parameter to `validateAuthCallback`. By default it is set to true so the change is non-breaking.

If it is set to false, all steps where cookies are checked are bypassed. Also, the session is loaded from the default offline session instead of from the cookie.

## Type of change

## Checklist

### DIFF
--- a/src/auth/oauth/oauth.ts
+++ b/src/auth/oauth/oauth.ts
@@ -120,9 +120,7 @@ const ShopifyOAuth = {
         );
       }
 
-      currentSession = await Context.SESSION_STORAGE.loadSession(
-        sessionCookie,
-      );
+      currentSession = await Context.SESSION_STORAGE.loadSession(sessionCookie);
     } else {
       currentSession = await Context.SESSION_STORAGE.loadSession(
         ShopifyOAuth.getOfflineSessionId(query.shop),
--- a/src/auth/oauth/oauth.ts
+++ b/src/auth/oauth/oauth.ts
@@ -156,7 +156,7 @@ const ShopifyOAuth = {
     const client = new HttpClient(currentSession.shop);
     const postResponse = await client.post(postParams);
 
-    if (isOnline) {
+    if (currentSession.isOnline) {
       const responseBody = postResponse.body as OnlineAccessResponse;
       const {access_token, scope, ...rest} = responseBody; // eslint-disable-line @typescript-eslint/naming-convention
       const sessionExpiration = new Date(
--- a/src/auth/oauth/oauth.ts
+++ b/src/auth/oauth/oauth.ts
@@ -105,7 +105,7 @@ const ShopifyOAuth = {
     request: http.IncomingMessage,
     response: http.ServerResponse,
     query: AuthQuery,
-    isOnline: boolean = true,
+    isOnline = true,
   ): Promise<SessionInterface> {
     Context.throwIfUninitialized();
     Context.throwIfPrivateApp('Cannot perform OAuth for private apps');
--- a/src/auth/oauth/oauth.ts
+++ b/src/auth/oauth/oauth.ts
@@ -156,7 +156,7 @@ const ShopifyOAuth = {
     const client = new HttpClient(currentSession.shop);
     const postResponse = await client.post(postParams);
 
-    if (currentSession.isOnline) {
+    if (isOnline) {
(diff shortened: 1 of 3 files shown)


src/auth/oauth/oauth.ts
```diff
@@ -188,12 +193,19 @@ const ShopifyOAuth = {
...
+        secure: true,
+      });
+
+      cookies.set(ShopifyOAuth.SESSION_COOKIE_NAME, currentSession.id, {
+        signed: true,
+        expires: Context.IS_EMBEDDED_APP ? new Date() : currentSession.expires,
+        sameSite: 'lax',
+        secure: true,
+      });
+    }
```
thecodepixi, line 208: Nit: Since this now only needs to happen for online sessions, can this be moved up to the bottom of the previous `currentSession.isOnline` check that starts on 160 and ends on 189? Could we put this between 188 and 189?

src/auth/oauth/test/oauth.test.ts
```diff
@@ -286,8 +288,16 @@ describe('validateAuthCallback', () => {
...
 
     fetchMock.mockResponse(JSON.stringify(successResponse));
-    await ShopifyOAuth.validateAuthCallback(req, res, testCallbackQuery);
-    session = await Context.SESSION_STORAGE.loadSession(cookies.id);
+    Cookies.prototype.set.mockImplementation(() => {
+      throw new Error('offline sessions should not depend on cookies');
+    });
+    Cookies.prototype.get.mockImplementation(() => {
+      throw new Error('offline sessions should not depend on cookies');
+    });
```
thecodepixi, line 296: Can you explain a little bit about what's happening here? Is this a new `Error` just for the context of this test?

thecodepixi (commented):
One nit and one testing related question, but this is looking really awesome and like it will solve the offline session/cookies issue nicely. Thanks for putting this together! Feel free to request a re-review from me whenever you have time to respond to the comments.

src/auth/oauth/oauth.ts
fjij: Yep, totally makes sense

src/auth/oauth/test/oauth.test.ts
fjij: Yea it is -- changed it to use `fail` from jasmine so it is clearer what is going on.

fjij:
I also removed the tests that were checking to see if the cookies were updated for offline sessions.

They didn't break before because I forgot to add the `isOnline = false` parameter to them.

mariusa:
Hi @thecodepixi 
Could this be committed please?

shoenseiwaso:
@thecodepixi it would still be great to get this in!

Given there are now conflicts, if we fix them, could we get a commitment from the Shopify side to get this reviewed and merged in a reasonable amount of time?

mkevinosullivan:
@shoenseiwaso Thanks for raising this to our attention.

The code does seem to make sense but wondering what specific scenario where cookies cannot be used for which you are trying to solve?  Depending on the scenario, there might be an alternative way to solve for it.

shoenseiwaso:
@mkevinosullivan would you be able to take another look at this? We would very much like to get this merged.

fjij:
@mkevinosullivan Rebased and updated changelog!

mkevinosullivan:
@fjij Would you mind rebasing again?  Sorry, lots of recent changes being merged in.

mkevinosullivan:
@fjij, @shoenseiwaso, sincere apologies on the delay on this one.  We've been going back-and-forth with our OAuth team in relation to a variety of OAuth items as part of a larger review and it appears that we're unable to commit this change.

We need to keep the cookie check in place as a countermeasure against [CSRF attacks against redirect-uri](https://datatracker.ietf.org/doc/html/rfc6819#section-4.4.1.8).  While we inherently know the value of the session id in the offline scenario and don't need the cookie value to look up the session details, the *presence* of the cookie is needed to confirm that the browser at the end of the OAuth flow is the same as the one that originated it to mitigate against this particular attack vector.

While we'll likely make some minor improvements to this component in the near future, we're unable to get rid of the cookie.

PR #298, opened 2022-02-05
Update rest.md

daviareias:
This is not working for anyone:
const session = await Shopify.Utils.loadCurrentSession(req, res);    

This one now works, I test with and without Redis:
const id = Shopify.Auth.getCurrentSessionId(req,res)
const session = await Shopify.Context.SESSION_STORAGE.loadSession(id)

Fixes #0000 

## Type of change

## Checklist

### DIFF
--- a/docs/usage/rest.md
+++ b/docs/usage/rest.md
@@ -44,7 +44,8 @@ We can run the code below in any endpoint where we have access to the request an
 
 ```ts
 // Load the current session to get the `accessToken`.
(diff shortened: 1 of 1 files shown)


paulomarg:
Thank you for contributing, but I don't think this is what we want to do - this is what [`loadCurrentSession`](https://github.com/Shopify/shopify-node-api/blob/main/src/utils/load-current-session.ts) already does, so there's no need to break it up into multiple statements 🙂

PR #301, opened 2022-02-11
Allow full paths in REST requests

paulomarg:

Currently, we can't handle requests like access scopes (`/admin/oauth/access_scopes.json`) or token access (`/admin/api_permissions/current.json`), because these resources use custom paths instead of the default `/admin/api/{version}/...`.

This means we can't easily reach those endpoints unless the code uses the base HTTP client that the REST one is built upon, which makes things unnecessarily complicated.

This PR solves that problem by allowing REST calls in the Admin API client to be made using a full path. Therefore, requests like

```js
const client = new RestClient(domain, 'token');
const response = await client.get({path: '/admin/some-path.json'});
```

work just like any regular request would.

## Type of change

## Checklist

### DIFF
--- a/src/clients/rest/test/rest_client.test.ts
+++ b/src/clients/rest/test/rest_client.test.ts
@@ -342,6 +342,36 @@ describe('REST client', () => {
       ShopifyErrors.MissingRequiredArgument,
     );
   });
+
+  it('allows paths with .json', async () => {
+    const client = new RestClient(domain, 'dummy-token');
+
+    fetchMock.mockResponseOnce(JSON.stringify(successResponse));
+
+    await expect(client.get({path: 'products.json'})).resolves.toEqual(
+      buildExpectedResponse(successResponse),
+    );
+    assertHttpRequest({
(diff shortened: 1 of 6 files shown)


src/clients/rest/test/rest_client.test.ts
```diff
@@ -342,6 +342,36 @@ describe('REST client', () => {
...
+
+    fetchMock.mockResponseOnce(JSON.stringify(successResponse));
+
+    await expect(client.get({path: '/admin/some-path.json'})).resolves.toEqual(
+      buildExpectedResponse(successResponse),
+    );
+    assertHttpRequest({
+      method: 'GET',
+      domain,
+      path: '/admin/api/unstable/admin/some-path.json',
```
mllemango, line 372: Shouldn't `/admin/some-path.json` be the full path here?

paulomarg: Yes, I got it wrong!

JaKXz (approved):

PR #310, opened 2022-03-01
Update supported Admin API versions

paulomarg:

Our supported Admin API versions were no longer up to date.

Adds the latest API versions, and drops unsupported ones. This is a breaking change in that using a non-supported version will create code errors (aside from any errors incurred by the silent API version upgrade).

## Type of change

## Checklist


src/webhooks/registry.ts
```diff
@@ -115,7 +115,7 @@ function isSuccess(
 
 // 2020-07 onwards
 function versionSupportsEndpointField() {
-  return ShopifyUtilities.versionCompatible(ApiVersion.July20);
+  return ShopifyUtilities.versionCompatible('2020-07' as unknown as ApiVersion);
```
eliasrumley, line 118: Since we're still using `July20` in our logic what was the intent of removing the following from our `ApiVersion` enum?

* `July20`
* `October20`
* `January21`

I would think it would be better to leave the original enum values, rather than using this double-cast trick to get around the type checker.

src/webhooks/registry.ts
```diff
@@ -256,7 +259,9 @@ const WebhooksRegistry: RegistryInterface = {
   },
 
   getHandler(topic: string): WebhookRegistryEntry | null {
-    return topic in WebhooksRegistry.webhookRegistry ? WebhooksRegistry.webhookRegistry[topic] : null;
+    return topic in WebhooksRegistry.webhookRegistry
+      ? WebhooksRegistry.webhookRegistry[topic]
+      : null;
```
eliasrumley, line 264: I believe we're doing this to avoid returning an `undefined` - would this simplify the code?
```suggestion
    return WebhooksRegistry.webhookRegistry[topic] ?? null;
```

src/webhooks/registry.ts
```diff
@@ -336,7 +341,8 @@ const WebhooksRegistry: RegistryInterface = {
           shop,
           deliveryMethod,
         };
-        const returnedRegister: RegisterReturn = await WebhooksRegistry.register(webhook);
+        const returnedRegister: RegisterReturn =
+          await WebhooksRegistry.register(webhook);
```
eliasrumley, line 345: TypeScript understands that `await Promise<T>` returns a `T`, so this should be able to depend on type inference
```suggestion
        const returnedRegister = await WebhooksRegistry.register(webhook);
```

src/webhooks/test/registry.test.ts
```diff
@@ -342,7 +344,7 @@ describe('ShopifyWebhooks.Registry.register', () => {
   it('throws if an eventbridge webhook is registered with an unsupported API version', async () => {
     expect(async () => {
       fetchMock.mockResponseOnce(JSON.stringify(webhookCheckEmptyResponse));
-      Context.API_VERSION = ApiVersion.April19;
+      Context.API_VERSION = '2019-04' as unknown as ApiVersion;
```
eliasrumley, line 347: Similar comments as before - double-casting to get around enums feels like an anti-pattern here. Happy to have a chat about what's going on here to see if there's another way to solve this.

src/webhooks/test/registry.test.ts
```diff
@@ -705,9 +785,18 @@ describe('ShopifyWebhooks.Registry.getTopics', () => {
...
+      webhookHandler: genericWebhookHandler,
+    });
+    ShopifyWebhooks.Registry.addHandler('PRODUCTS_CREATE', {
+      path: '/webhooks',
+      webhookHandler: genericWebhookHandler,
+    });
+    expect(ShopifyWebhooks.Registry.getTopics()).toStrictEqual([
+      'PRODUCTS',
+      'PRODUCTS_CREATE',
+    ]);
```
eliasrumley, line 799: These tests have a massive amount of copy-paste in them, and are too verbose to really understand. I realize that most of the changes here are formatting, but there's a part of me that would almost like to see it re-factored to use proper functions; this way, the tests could read like a spec.

As an example, here's what two of these tests could look like:

```typescript
it('adds multiple handlers with duplicates', async () => {
  addHandler('PRODUCTS');
  addHandlers({
    PRODUCTS_CREATE: '/webhooks',
    PRODUCTS: '/newerpath'
  });
  assertRegistryKeysLength(2);
  assertRegistryKeys(['PRODUCTS_CREATE', 'PRODUCTS'])
  assertRegisteredPath('PRODUCTS', '/newerpath');
});

it('adds two handlers and gets them', async () => {
  addHandler('PRODUCTS');
  addHandler('PRODUCTS_CREATE');
  assertTopics([
    'PRODUCTS',
    'PRODUCTS_CREATE'
  ]);
});

// there's a way to type this, I just don't know it off the top of my head

function addHandlers(hash) {
  const finalHash = {};

  for(let key in hash) {
    if(hash.hasOwnProperty(key)) {
      finalHash[key] = {
        path: hash[key],
        webhookHandler: genericWebhookHandler
      }
    }
  }

  await ShopifyWebhooks.Registry.addHandlers(finalHash);
}

function addHandler(type: string) {
  await ShopifyWebhooks.Registry.addHandler(type, {
    path: '/webhooks',
    webhookHandler: genericWebhookHandler,
  });
}

function assertRegistryKeysLength(expected: number) {
  expect(Object.keys(ShopifyWebhooks.Registry.webhookRegistry)).toHaveLength(expected);
}

function assertRegisteredPath(type: string, expected: string) {
  expected(ShopifyWebHooks.Registry.webhookRegistry[type].path).toBe(expected);
}

function assertRegistryKeys(expected: string[]) {
  expected(Object.keys(ShopifyWebhooks.Registry.webhookRegistry).sort()).toEqual(expected.sort());
}

function assertTopics(expected: string[]) {
  expect(ShopifyWebhooks.Registry.getTopics()).toStrictEqual(expected);
}
```

Thoughts?

src/webhooks/registry.ts
paulomarg: That's a fair point - I think we should actually just fully drop this function - `2020-07` is no longer supported, so this branch should never run in a real world setting.

src/webhooks/test/registry.test.ts
paulomarg: As per above, all of this code can just go away.

eliasrumley (approved):
:+1: Awesome, looks great!

PR #311, opened 2022-03-01
Add base rest resource

paulomarg:
This PR adds support for a base resource that can be used to create objects representing the Admin API. In the process of adding it, we noticed that the tests were not ideally placed, and that we could improve how we asserted HTTP requests.

## 🎩 / review guide:

This file is where the core changes are:
https://github.com/Shopify/shopify-node-api/pull/311/files#diff-5e6c775a84b8590f3a54b3743699d17a779b9467a560b5f1df12d427538bdc70

There are many prettier/eslint/test formatting changes courtesy of @JaKXz's pedantry :D


src/rest_resources/__tests__/base.test.ts
```diff
@@ -0,0 +1,396 @@
+import {Session} from '../../auth/session';
+import {RestResourceRequestError} from '../../error';
+
+import FakeResource from './fake_resource';
+import FakeResourceWithCustomPrefix from './fake_resource_with_custom_prefix';
+
+describe('Base REST resource', () => {
```
eliasrumley, line 7: I think this class could benefit from reduced duplication in the different test cases.

I don't think extracting an object like in the prod code above would make sense, but helper functions could go a long way.

src/rest_resources/base.ts
```diff
@@ -0,0 +1,363 @@
...
+    });
+
+    this.NEXT_PAGE_INFO = undefined;
+    this.PREV_PAGE_INFO = undefined;
+    if (response.pageInfo?.nextPage) {
+      this.NEXT_PAGE_INFO = response.pageInfo.nextPage;
+    }
+    if (response.pageInfo?.prevPage) {
+      this.PREV_PAGE_INFO = response.pageInfo.prevPage;
+    }
```
eliasrumley, line 92: Ternary to reduce lines of code?
```typescript
this.NEXT_PAGE_INFO = response.pageInfo?.nextPage ?? undefined;
this.PREV_PAGE_INFO = response.pageInfo?.prevPage ?? undefined;
```

src/rest_resources/base.ts
```diff
@@ -0,0 +1,363 @@
...
+          });
+          break;
+        case 'delete':
+          response = await client.delete({path, query: cleanParams});
+          break;
+        default:
+          throw new Error('Unrecognized HTTP method');
+      }
+
+      return response;
```
eliasrumley, line 148: 1. Could shrink code and improve self-documentation by using `return` in `switch` statement.
2. Error response would be more useful if it stated what the unrecognised HTTP method was

```suggestion
      switch (http_method) {
        case 'get':
          return await client.get({path, query: cleanParams});
        case 'post':
          return await client.post({
            path,
            query: cleanParams,
            data: body!,
            type: DataType.JSON,
          });
        case 'put':
          return await client.put({
            path,
            query: cleanParams,
            data: body!,
            type: DataType.JSON,
          });
        case 'delete':
          return await client.delete({path, query: cleanParams});
        default:
          throw new Error('Unrecognized HTTP method "' + http_method + '"');
      }
```

eliasrumley (commented):
Lots of good work done here. I do have some feedback that could reduce a bit of code and/or improve the readability, let me know your thoughts! :smile:

src/clients/http_client/http_client.ts
```diff
@@ -116,11 +116,33 @@ class HttpClient {
...
+          );
+        } else {
+          processedQuery[key] = value;
+        }
+      });
+
+      queryString = `?${querystring.stringify(
+        processedQuery as ParsedUrlQueryInput,
+      )}`;
+    }
```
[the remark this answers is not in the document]
JaKXz: Thanks!

src/rest_resources/__tests__/base.test.ts
paulomarg: I kind of like it the way it is, because the tests are essentially doing the Arrange -> Act -> Assert flow, and each of the calls there are a single call to the functions that do their part.

src/rest_resources/base.ts
JaKXz: thanks!

JaKXz (approved):

eliasrumley (approved):

PR #312, opened 2022-03-02
Stop sending request response from the GraphQL proxy

paulomarg:

We noticed that the way the GraphQL proxy currently works is inconsistent with how the webhook `process` method works - while the webhook method returned the result of its operation, the proxy was triggering the actual response to be sent to the client, which was quite confusing.

Changing the GraphQL proxy so it behaves consistently by returning the response of the queried request to the app, rather than taking over and responding to the request itself.

This allows apps to take the body of a proxied request and 1) respond however they prefer; 2) run any code they need to after sending the request.

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


tolgap:
This seems to be a breaking change. But what actually breaks? The changelog doesn't mention what to do to if you upgrade.

PR #319, opened 2022-03-04
Fix shopify response is not json

zookcorp:
bug fix -  when response status 404 but response body is not json

### DIFF
--- a/src/clients/http_client/http_client.ts
+++ b/src/clients/http_client/http_client.ts
@@ -180,8 +180,14 @@ class HttpClient {
   ): Promise<RequestReturn> {
     return fetch(url, options)
       .then(async (response: Response) => {
-        const body = await response.json();
-
+        let body;
+      
+        try {
+          body = await response.json();
+        } catch {
+          body = {};
+        }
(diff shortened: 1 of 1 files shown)


src/clients/http_client/http_client.ts
```diff
@@ -180,7 +180,13 @@ class HttpClient {
...
     return fetch(url, options)
       .then(async (response: Response) => {
-        const body = await response.json();
+        let body;
+
+        try {
+          body = await response.json();
+        } catch {
+          body = {};
+        }
```
JaKXz, line 189: Here's a suggestion to simplify this :) 

```suggestion
        const body = await response.json().catch(() => ({}));
```

Although now that I'm looking at this file @paulomarg we should refactor this. Thanks @zookcorp!

zookcorp:
> Thanks for contributing @zookcorp! I think the change makes sense, if you could fix the linting errors I'll go ahead and merge this 🙂

Sorry, I edited it directly on git and didn't notice to linting. i fixed it. 😋

PR #345, opened 2022-03-29
[WIP] Add tests using real (not mocked) node/cf-worker adapters

mkevinosullivan:

With the abstraction of the network calls in order to make the library isomorphic, we want to add tests that actually validate the real network calls, in both Node and Cloudflare Workers (using `miniflare`)

- Adds utilities to
  - determine if we're running tests by checking for `E2ETESTS` environment variable (node) or global binding (CF worker)
  - return a platform string, either `Node <version>` or `Cloudflare Worker, for use in the User-Agent string
- Removes dependency on Buffer (Node-specific)
- Adds setting of `crypto` in cf-worker adapter
- Bypasses URL validator check if running network tests (`E2ETESTS` set)

NB: contents of `experiments` folder will be deleted

#### Node
```shell
yarn e2etests:node
```
Using `concurrently`, this will call
- `yarn e2etests:http_server`, which runs a basic node HTTP server on port `3000` that responds to the test requests
- `yarn e2etests:node_client`, which runs the `jest`-based test suite `src/clients/http_client/__e2etests__/http_client-node.test.ts`, which is a copy of `src/clients/http_client/http_client.test.ts` but uses the node adapter instead of the mock adapter.

#### Cloudflare Worker
```shell
yarn e2etests:cf_worker
```
Using `concurrently`, this will call
- `yarn e2etests:http_server`, which runs a basic node HTTP server on port `3000` that responds to the test requests
- `yarn e2etests:serve:cf_worker`, which first runs `yarn e2etests:build:cf_worker` to bundle the js using `rollup` and then runs it using `miniflare`
  - This creates a Cloudflare worker (a _server_, on port `8787`) that represents a back-end that communicates as a _client_ with the HTTP server
  - The tests themselves are non-`jest` implementations of 23 of the 25 tests in the node test suite.
- `sleep 2 && curl http://localhost:8787` - wait 2 seconds (for the build and processes above to be ready) and then send a `get` request to the CF worker to initiate the tests

**NB**: because `miniflare` runs a script as a server and has no `process.exit()` like Node to gracefully exit (at least none that I could find!), the `concurrently` command is run with `--kill-others` option, which kills all the other processes once the tests have finished.  Hence, the tests always "fail" with error code 1, even when the actual tests themselves pass.

### Running all e2e tests
```shell
yarn e2etests:test
```


mkevinosullivan:
There's another alternative structure (preferred) that I've added, which overcomes the major deficiencies of how the CF Worker tests are executed as described in the original description:

#### Cloudflare Worker (alternative
```shell
yarn e2etests:cf_worker2
```
Using `concurrently`, this will call
- `yarn e2etests:http_server`, which runs a basic node HTTP server on port `3000` that responds to the test requests
- `yarn e2etests:serve2:cf_worker`, which runs `yarn node --experimental-vm-modules src/clients/http_client/__e2etests__/experiments/miniflare_test.mjs`
  - `miniflare_test.mjs` includes the build command (`e2etests:build:cf_worker`)
  - Then it runs the bundled script _inside_ a `Miniflare()` object within a Node script

Benefits:
- Being inside a Node script means that the triggering `get` can be done internally as an event dispatch, removing the need for the external `curl` command
- Being inside a Node script means that the exit code can be set to properly indicate pass/fail of the test suite

Current challenges:
- Passing a `E2ETESTS` environment variable or binding into the script via the `Miniflare()` object isn't working as I expected - I've temporarily changed it to use `globalThis.MINIFLARE` to determine if the network tests are being run or not, which is not quite ideal.

src/clients/http_client/http_client.ts
```diff
@@ -116,8 +119,10 @@ class HttpClient {
             break;
         }
         headers = {
+          /* eslint-disable @typescript-eslint/naming-convention */
           'Content-Type': type,
-          'Content-Length': Buffer.byteLength(body as string).toString(),
+          'Content-Length': (body as string).length.toString(),
```
mkevinosullivan, line 124: `Buffer.byteLength` is from Node and it's not available in Cloudflare Worker (another issue caught by the tests!)

Since all the cases in the `switch` preceding this ensure that `body` is set to a `string`, I think it's fairly safe to use the string`.length` method to set the `Content-Length` ... thoughts? 🤔

surma: No, not quite, because JavaScript strings are UCS-2 encoded, but strings will be UTF-8 encoded when sent over the wire by most (all?) web server implementations. Those can have different lengths for the same string. For example:

```js 
"👪".length // => 2
new TextEncoder("utf8").encode("👪").length // => 4
```

I’d recommend we use the latter to set `Content-Length`.

src/utils/network-tests.ts
```diff
@@ -0,0 +1,18 @@
...
+ */
+export default function runningNetworkTests(): boolean {
+  if (typeof process !== 'undefined') {
+    // process variable exists => NodeJS
+    // eslint-disable-next-line no-process-env
+    return process.env.E2ETESTS !== 'undefined' && process.env.E2ETESTS === '1';
+  }
+  if (typeof globalThis !== 'undefined') {
+    // not in node, therefore CF worker or miniflare
+    return ( (globalThis as any).MINIFLARE
```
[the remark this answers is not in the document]
surma: Yeah, I agree. I think we’d rather have another ENV variable that needs to be set than changing the behavior just because the developer is using a specific runtime and than gets divergent behavior.

PR #346, opened 2022-03-31
Add regex to shop validator to support Spin urls

carrotderek:

Currently the shop validator will fail for apps on Spin. 

PR is a revival of a previously [closed one](https://github.com/Shopify/shopify-node-api/pull/243) that extends the regex in the shop validator to include Spin-like urls (ie. `shop1.shopify.<workspace>.<namespace>.<region>.spin.dev`).

## Type of change

## Checklist

  - Is this a noteworthy enough change to document in the changelog?
  - Nothing really to document

### DIFF
--- a/src/utils/shop-validator.ts
+++ b/src/utils/shop-validator.ts
@@ -4,6 +4,7 @@
  * @param shop Shop url: {shop}.myshopify.com
  */
 export default function validateShop(shop: string): boolean {
-  const shopUrlRegex = /^[a-zA-Z0-9][a-zA-Z0-9-]*\.myshopify\.(com|io)[/]*$/;
+  const shopUrlRegex =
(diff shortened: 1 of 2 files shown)


src/utils/shop-validator.ts
```diff
@@ -4,6 +4,7 @@
  * @param shop Shop url: {shop}.myshopify.com
  */
 export default function validateShop(shop: string): boolean {
-  const shopUrlRegex = /^[a-zA-Z0-9][a-zA-Z0-9-]*\.myshopify\.(com|io)[/]*$/;
+  const shopUrlRegex =
+    /(^[a-zA-Z0-9][a-zA-Z0-9-]*\.myshopify\.(com|io)[/]*$|^[a-zA-Z0-9][a-zA-Z0-9-]*\.shopify\.[a-zA-Z0-9][a-zA-Z0-9-]*\.[a-zA-Z0-9][a-zA-Z0-9-]*\.(us|asia|eu)\.spin.dev[/]*$)/;
```
paulomarg, line 8: The regex works for me but it's a bit hard to read in this format. Do you think it would be easier to read if we put each of the individual components into strings and then just `|` them in the final regex?

carrotderek: For sure I can split them up into different variables 👍🏼

carrotderek:
Closing in favor of https://github.com/Shopify/shopify-node-api/pull/353

PR #347, opened 2022-04-01
Update REST resources after 2022-04 release

paulomarg:
This is just updating the auto-generated resources following the release of the `2022-04` API version (with some bonus improvements to the metafields resources).


mkevinosullivan (approved):
I haven't tested this but I'm assuming it's all OK!

PR #353, opened 2022-04-05
Remove shop validator

carrotderek:

There were proposed changes to the shop validator in https://github.com/Shopify/shopify-node-api/pull/346 to be able to support Spin urls. From discussions within the team, it was raised whether the validator adds any real value.

PR removes `validateShop` in wherever it is used:
 - Validating callback query during OAuth authentication
 - Decoding of  JWT session token
 - Instantiation of http client

We are keeping the function and test around albeit with a deprecation notice.

## Type of change

## Checklist

  - Leaving this for reviewers if this is a noteworthy change


src/utils/index.ts
```diff
@@ -23,7 +22,6 @@ const ShopifyUtils = {
   safeCompare,
   storeSession,
   validateHmac,
-  validateShop,
```
paulomarg, line 26: I think this all makes sense, but if we stop exporting the function from here, this becomes a breaking change because it changes the lib's external API. We could probably keep the function around and `console.warn` until the next major release, just so we don't tack on a major release right after the last one?

carrotderek: Good call, I brought the function + test back and added the deprecation notice.

paulomarg (approved):

PR #366, opened 2022-04-16
[Feature] Add support for http2

maxfrigge:
The cookie related methods were not compatible with http2.

The current version does not support the http2 types of the request/response Node objects. This makes it impossible to use the package in a http2 server written in Typescript, unless you are OK with ignoring the type errors.

Replace the [cookies package](https://www.npmjs.com/package/cookies) (which also writes the response header) with the [cookie package](https://www.npmjs.com/package/cookie) (which just parses cookie strings). This requires to manually set and read the cookie header, but gives greater control and allows to also work with the http2 request/response interfaces.

One thing to note is that the new cookie package won't sign cookies. My understanding is that this is not necessary when using `secure` and `httponly` cookies, but **should be confirmed be a security expert**.  This change could potentially break existing cookies, but should "just" cause a new login flow, which then updates to an unsigned cookie. 

I have added `@ts-ignore` comments to ignore unused variable errors for the arguments of `getCookieSessionId` and `setCookieSessionId` in order to not introduce a breaking change in the api surface.

## Type of change

## Checklist


maxfrigge:
@paulomarg I have created this new PR after #232 was closed.

paulomarg:
Hey, sorry for the delay in responding. In the end, we're not even using `http` directly any more so this doesn't really apply. Thank you for your patience and for contributing to our repo!

PR #372, opened 2022-04-25
Make SQLite the default SessionStorage

surma:
I also had to update TypeScript because many of the database-provided types don’t work with TS3.


src/auth/session/session-utils.ts
```diff
@@ -0,0 +1,44 @@
+import type {SessionInterface} from './types';
```
surma, line 1: This is old code just moved to a separate file so that the linter stops complaining.

PR #376, opened 2022-04-27
Add some additional doc on using GraphQL variables

nickwesselman:

I was trying to use GraphQL variables, and it wasn't clear based on docs what the structure of `data` should be in that case.

Adds this information, and an additional example to the doc which shows it in-use.

## Type of change

Doc-only

## Checklist

Doc-only, so I did none of these, but let me know if you still need a changelog entry.


paulomarg (approved):
This is great, thank you!

teddyhwang:
@surma looks like we have some flakiness with the session storage tests?

surma:
Fixed it! Gonna go ahead and merge :)

PR #380, opened 2022-05-04
Adds support for read-only attributes to serialize

mkevinosullivan:

To align the behaviour of `serialize` with that of the Ruby `to_hash` method, as updated by https://github.com/Shopify/shopify_api/pull/941

Previously, `serialize` excluded read-only attributes, since it was assumed it was just being used by `save` method.

Now, default is that `serialize` returns all attributes, and an optional (`saving=`)`true` parameter will exclude them when used inside `save` method.

## Type of change

## Checklist

### DIFF
--- a/src/base-rest-resource.ts
+++ b/src/base-rest-resource.ts
@@ -256,7 +256,7 @@ class Base {
     const {PRIMARY_KEY, NAME} = this.resource();
     const method = this[PRIMARY_KEY] ? 'put' : 'post';
 
-    const data = this.serialize();
+    const data = this.serialize(true);
 
     const response = await this.resource().request({
       http_method: method,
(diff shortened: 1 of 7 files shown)


src/base-rest-resource.ts
```diff
@@ -288,25 +288,26 @@ class Base {
...
+          READ_ONLY_ATTRIBUTES.includes(attribute))
       ) {
         return acc;
       }
 
       if (attribute in HAS_MANY && value) {
         acc[attribute] = value.reduce(
           (attrAcc: Body, entry: Base) =>
-            attrAcc.concat(entry.serialize ? entry.serialize() : entry),
+            attrAcc.concat(entry.serialize ? entry.serialize(saving) : entry),
```
paulomarg, line 306: We should probably also do the thing here where if `!entry.serialize`, we create a new instance so that we can `serialize` it.

paulomarg (commented):
LGTM, just need to take it one step further IMO!

paulomarg (approved):
🚢

PR #385, opened 2022-05-16
Don't assume the host URI scheme is https

pepicrft:
We'll allow loading apps from localhost through `http` in the upcoming CLI release. To make it work, we need to adjust the logic in this package to not assume that the URI scheme is `https`. 

I'm removing the `https` assumption and allowing the consumer of the package to configure the URI scheme through the new context attribute, `HOST_SCHEME`. To make the change backward-compatible, I'm defaulting to `https` when no value is provided.

## Type of change

## Checklist

### DIFF
--- a/src/base-types.ts
+++ b/src/base-types.ts
@@ -6,6 +6,7 @@ export interface ContextParams {
   API_SECRET_KEY: string;
   SCOPES: string[] | AuthScopes;
   HOST_NAME: string;
+  HOST_SCHEME?: string;
   API_VERSION: ApiVersion;
   IS_EMBEDDED_APP: boolean;
   IS_PRIVATE_APP?: boolean;
--- a/src/context.ts
+++ b/src/context.ts
@@ -7,6 +7,7 @@ import {ApiVersion, ContextParams} from './base-types';
 import {AuthScopes} from './auth/scopes';
 
 interface ContextInterface extends ContextParams {
+  HOST_SCHEME: string;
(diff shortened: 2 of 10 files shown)


src/base-types.ts
```diff
@@ -6,6 +6,7 @@ export interface ContextParams {
   API_SECRET_KEY: string;
   SCOPES: string[] | AuthScopes;
   HOST_NAME: string;
+  HOST_SCHEME?: string;
```
shauns, line 9: Do we want this to be optional? Or is there a separate type for input (where it is optional) and after applying defaults (where it will be set)?

gonzaloriestra (approved):
👌

src/base-types.ts
pepicrft: I'm making it optional to prevent the change from breaking. Otherwise existing users that update the package will get Typescript errors because the type-checker will expect that value to be passed.

src/context.ts
```diff
@@ -82,6 +84,10 @@ const Context: ContextInterface = {
     this.IS_EMBEDDED_APP = params.IS_EMBEDDED_APP;
     this.IS_PRIVATE_APP = params.IS_PRIVATE_APP;
 
+    if (params.HOST_SCHEME) {
+      this.HOST_SCHEME = params.HOST_SCHEME;
```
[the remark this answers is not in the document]
pepicrft: We definitively should. Let me add it.

PR #393, opened 2022-05-26
Fix sqlite creation

### LINKED ISSUE #389: Empty `database.sqlite` being created.
# Issue summary

An empty `database.sqlite` is being created the moment `Shopify.Context.initialize({})` is called.

## Expected behavior

No `database.sqlite` files should be created, or should have a purpose of creation.

## Actual behavior

A `database.sqlite` with the heads `id, shop, state, isOnline, scope, accessToken` is being created in the root folder of the `Shopify.Context.initialize({})`. To put things in perspective, in a boilerplate [repo](https://github.com/kinngh/shopify-node-express-mongodb-app), the `database.sqlite` folder is being created in the `server/`, which means it creates the database in the root of where the init is called.

A barebones repo has been created here: [https://github.com/kinngh/api-issue-barebone](https://github.com/kinngh/api-issue-barebone)

- Run `npm install` followed by a `node index.js` to see the file being created.
- This was created to isloate the API that the moment it's called, it creates a sqlite database.

## Reduced test case

None

---

## Checklist

surma:

Fixes #389

Even if users overwrite the `SESSION_STORAGE` in the context, the default value still creates a `database.sqlite` file as a side-effect. This should not happen.

To avoid a breaking change, I have augmented the Context. The `SESSION_STORAGE` now accepts `SessionStorage`, but also _functions_ that return a `SessionStorage`. This way, the side effects of any session storage are delayed until they are accessed for the first time.
(This is achieved through a getter/setter combination).


src/context.ts
```diff
@@ -33,15 +33,29 @@ const dbFile = path.join(
...
   API_KEY: '',
   API_SECRET_KEY: '',
   SCOPES: new AuthScopes([]),
   HOST_NAME: '',
   API_VERSION: ApiVersion.Unstable,
   IS_EMBEDDED_APP: true,
   IS_PRIVATE_APP: false,
-  SESSION_STORAGE: new SQLiteSessionStorage(dbFile),
+
+  SESSION_STORAGE_: () => new SQLiteSessionStorage(dbFile),
```
paulomarg, line 49: What if instead of doing this here and adding the getter / setter, we only changed `initialize` (i.e. `ContextParams`) to take in a function and default this to a null value (or memory storage), defaulting to SQLite in initialize?

src/context.ts
```diff
@@ -33,15 +33,29 @@ const dbFile = path.join(
...
+  get SESSION_STORAGE(): SessionStorage {
+    if(typeof this.SESSION_STORAGE_ === 'function') {
+      this.SESSION_STORAGE_ = this.SESSION_STORAGE_();
```
paulomarg, line 52: I wonder if it would make sense to not call the function every time, but call it once on initialization and memoize it. We could let it be an async function, and then the storage class can e.g. connect to a remote DB server when it's created, and we await it here.

paulomarg (commented):
Good stuff! I think this works as is, but I had a couple of thoughts 🙂

src/context.ts
surma: That is probably the better suggestion, yes. I’ll change that

src/context.ts
paulomarg: Right, brilliant!

surma:
@paulomarg Your suggestion was the much better solution. Fixed that up, PTAL :)

paulomarg (approved):
🔥

kinngh:
@surma can we please have a version bump to `3.1.1`?

PR #394, opened 2022-05-27
Augment session data

surma:
Augments the database adapters to store the expiry date and the user ID of the session token.

Note: I am actively omitting the users name and email address etc to avoid storing PII without the user potential realizing that.

### DIFF
--- a/src/auth/session/session-utils.ts
+++ b/src/auth/session/session-utils.ts
@@ -42,3 +42,14 @@ export function sessionEntries(
 ): [string, string | number][] {
   return Object.entries(session).filter(([key]) => includedKeys.includes(key));
 }
+
+export function sessionEqual(
+  sessionA: SessionInterface | undefined,
+  sessionB: SessionInterface | undefined,
+): boolean {
+  if (!sessionA) return false;
+  if (!sessionB) return false;
+  const copyA = sessionFromEntries(sessionEntries(sessionA));
+  const copyB = sessionFromEntries(sessionEntries(sessionB));
+  return JSON.stringify(copyA) === JSON.stringify(copyB);
+}
--- a/src/auth/session/session-utils.ts
+++ b/src/auth/session/session-utils.ts
@@ -24,7 +24,7 @@ export function sessionFromEntries(
       })
       // Sanitize values
       .map(([key, value]) => {
-        switch(key) {
+        switch (key) {
           case 'isOnline':
             if (typeof value === 'string') {
               return [key, value.toString().toLowerCase() === 'true'];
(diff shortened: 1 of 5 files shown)


src/auth/session/session-utils.ts
```diff
@@ -9,23 +9,49 @@ export function sessionFromEntries(
...
+            if (typeof value === 'string') {
+              return [key, value.toString().toLowerCase() === 'true'];
+            } else if (typeof value === 'number') {
+              return [key, Boolean(value)];
+            }
+            return [key, value];
+          case 'scope':
+            return [key, value.toString()];
+          case 'expires':
+            return [key, new Date(Number(value) * 1000)];
```
paulomarg, line 38: This is potentially null, do we need to handle that here?

src/auth/session/session-utils.ts
```diff
@@ -36,9 +62,36 @@ const includedKeys = [
...
 ): [string, string | number][] {
-  return Object.entries(session).filter(([key]) => includedKeys.includes(key));
+  return (
+    Object.entries(session)
+      .filter(([key]) => includedKeys.includes(key))
+      // Prepare values for db storage
+      .map(([key, value]) => {
+        switch (key) {
+          case 'expires':
+            return [key, Math.floor(value.getTime() / 1000)];
```
paulomarg, line 78: Same thing here about null expiration times

src/auth/session/storage/__tests__/battery-of-tests.ts
```diff
@@ -20,13 +25,77 @@ export function batteryOfTests(storageFactory: () => Promise<SessionStorage>) {
...
+    );
+
+    await expect(storage.deleteSession(sessionId)).resolves.toBe(true);
+    await expect(storage.loadSession(sessionId)).resolves.toBeUndefined();
+
+    // Deleting a non-existing session should work
+    await expect(storage.deleteSession(sessionId)).resolves.toBe(true);
+  });
+
+  it('can store and delete offline sessions with user info', async () => {
```
paulomarg, line 53: Technically, offline sessions won't ever have user info, but I don't think we need any special code to block saving the sessions in that case.

src/auth/session/session-utils.ts
surma: Good catch! Added test and fix

src/auth/session/session-utils.ts
```diff
@@ -9,23 +9,49 @@ export function sessionFromEntries(
...
+            if (typeof value === 'string') {
+              return [key, value.toString().toLowerCase() === 'true'];
+            } else if (typeof value === 'number') {
+              return [key, Boolean(value)];
+            }
+            return [key, value];
+          case 'scope':
+            return [key, value.toString()];
+          case 'expires':
+            return [key, value ? new Date(Number(value) * 1000) : undefined];
```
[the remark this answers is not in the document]
paulomarg: Makes sense to me!

PR #398, opened 2022-05-31
allow for non-json responses in the body

### LINKED ISSUE #257: Transient shopify error (for html response instead of json) is thrown as HttpRequestError instead of HttpRetriableError

JaKXz:
- refactors HttpClient#doRequest method to pull out nested .then callback
- based on PR / closes #319
- fixes #257 

Co-Authored-By: Paulo Margarido <64600052+paulomarg@users.noreply.github.com>

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+- Fixes [#257](https://github.com/Shopify/shopify-node-api/issues/257)
+  - Properly parse a non-json HTTP response
 
 ## [3.1.0] - 2022-05-16
 


paulomarg (approved):
🚢

PR #409, opened 2022-06-20
Add upcoming July22 API version

paulomarg:

Version `2022-07` of the API is coming up, and we should support it in the library.

Adding the version, but not changing any defaults or removing older versions while they're still supported.

## Type of change

## Checklist

### DIFF
--- a/src/base-types.ts
+++ b/src/base-types.ts
@@ -21,6 +21,7 @@ export enum ApiVersion {
   October21 = '2021-10',
   January22 = '2022-01',
   April22 = '2022-04',
+  July22 = '2022-07',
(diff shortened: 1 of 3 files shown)


src/base-types.ts
```diff
@@ -21,6 +21,7 @@ export enum ApiVersion {
   October21 = '2021-10',
   January22 = '2022-01',
   April22 = '2022-04',
+  July = '2022-07',
```
teddyhwang, line 24:
```suggestion
  July22 = '2022-07',
```

teddyhwang (approved):

PR #415, opened 2022-06-29
Add 2022-07, remove 2021-07 REST Resources

mkevinosullivan:

To support the quarterly addition/deprecation of API versions.

Does the Node part of https://github.com/Shopify/first-party-library-planning/issues/372

- Adds REST resources for `2022-07`
- Removes REST resources for `2021-07`
- Adds new `LATEST_API_VERSION` constant and points it to `2022-07`

## Type of change

## Checklist


meifif (approved):

meifif:
Great job, any timeline on realising?

src/index.ts
```diff
@@ -5,9 +5,11 @@ import ShopifySession from './auth/session';
 import ShopifyClients from './clients';
 import ShopifyUtils from './utils';
 import ShopifyWebhooks from './webhooks';
+import {LATEST_API_VERSION} from './base-types';
 
 export const Shopify = {
   Context,
+  LATEST_API_VERSION,
```
paulomarg, line 12: I haven't tested this, but I suspect we may already be exporting this in line 22 below:

```ts
export * from './types';
```

If not, I think it might be best to export it from there.

paulomarg (approved):
Just a nit, but LGTM otherwise!

PR #418, opened 2022-07-08
Add optional deleteSessions, findSessionsByShop

mkevinosullivan:

To support using the session database as a means to track app installations, adding optional methods to search for sessions by shop and deleting all sessions by a range of ids (e.g., for a shop, or as a garbage collector, if desired).

Adding optional `findSessionsByShop` and `deleteSessions` interface methods to `SessionStorage` and the corresponding implementations in the various DB adapters.

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+- Add optional new methods `deleteSession` and `findSessionsByShop` to `SessionStorage`, with the corresponding implementations for the various session storage adapters [#418](https://github.com/Shopify/shopify-api-node/pull/418)
+
 ## [4.1.0] - 2022-07-14
 
 - Add new method to construct the host app URL [#419](https://github.com/Shopify/shopify-api-node/pull/419)
--- a/docs/issues.md
+++ b/docs/issues.md
@@ -4,8 +4,14 @@ By following this guide, you will have a fully functional Shopify app. However, 
 
 ## Notes on session handling
 
-Before you start writing your application, please note that the Shopify library stores some information for OAuth in sessions. Since each application may choose a different strategy to store information, the library cannot dictate any specific storage strategy. By default, `Shopify.Context` is initialized with `MemorySessionStorage`, which will enable you to start developing your app by storing sessions in memory. You can quickly get your app set up using it, but please keep the following in mind.
+Before you start writing your application, please note that the Shopify library stores some information for OAuth in sessions. Since each application may choose a different strategy to store information, the library cannot dictate any specific storage strategy. By default, `Shopify.Context` is initialized with `SQLiteSessionStorage`, which will enable you to start developing your app by storing sessions using the file-based SQLite package. You can quickly get your app set up using it, but please keep the following in mind.
 
-`MemorySessionStorage` is **purposely** designed to be a single-process, development-only solution. It **will leak** memory in most cases and delete all sessions when your app restarts. You should **never** use it in production apps. In order to use a `CustomSessionStorage` solution in your production app, you can reference our [usage example with redis](usage/customsessions.md) to get started.
+`MemorySessionStorage` is **purposely** designed to be a single-process, development-only solution. It **will leak** memory in most cases and delete all sessions when your app restarts. You should **never** use it in production apps.
+
+`SQLiteSessionStorage` (the library default) is designed to be a single-process solution and *may* be sufficient for your production needs depending on your applications design and needs.
+
+The other storage adapters (`MongoDBSessionStorage`, `MySQLSessionStorage`, `PostgreSQLSessionStorage`, `RedisSessionStorage`) cover a variety of production-grade storage options.
+
+If you wish to have an alternative storage solution, you must use a `CustomSessionStorage` solution in your production app- you can reference our [usage example with redis](usage/customsessions.md) to get started.
 
 [Back to guide index](README.md)
--- a/docs/usage/customsessions.md
+++ b/docs/usage/customsessions.md
@@ -1,10 +1,16 @@
 # Create a `CustomSessionStorage` solution
 
-This library comes with two session management options: `MemorySessionStorage` and `CustomSessionStorage`.
+This library comes with various session management options:
 
-`MemorySessionStorage` exists as an option to help you get started developing your apps as quickly as possible, and is the default storage option on `Shopify.Context`. It's perfect for working in your development and testing environments. However, this storage solution is not meant to be used in production [due to its limitations](../issues.md).
+- `CustomSessionStorage` - to allow for a custom session storage solution (see below for details).
+- `MemorySessionStorage` - uses memory exists as an option to help you get started developing your apps as quickly as possible. It's perfect for working in your development and testing environments. However, this storage solution is **not** meant to be used in production [due to its limitations](../issues.md).
+- `MongoDBSessionStorage`
+- `MySQLSessionStorage`
+- `PostgreSQLSessionStorage`
+- `RedisSessionStorage`
+- `SQLiteSessionStorage` - uses the file-based SQLite package, and is the default storage option on `Shopify.Context`.
 
-When you're ready to deploy your app and run it in production, you'll need to set up a `CustomSessionStorage`, which you can then use in initializing your `Shopify.Context`. The `CustomSessionStorage` class expects to be initialized with three callbacks that link to your chosen storage solution and map to the `storeSession`, `loadSession`, and `deleteSession` methods on the class.
+If you wish to you an alternative session storage solution for production, you'll need to set up a `CustomSessionStorage`, which you can then use in initializing your `Shopify.Context`. The `CustomSessionStorage` class expects to be initialized with the following three mandatory callbacks that link to your chosen storage solution and map to the `storeSession`, `loadSession`, and `deleteSession` methods on the class.
 
 ## Callback methods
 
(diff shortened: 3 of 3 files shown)


src/auth/session/storage/custom.ts
```diff
@@ -5,20 +5,32 @@ import * as ShopifyErrors from '../../../error';
...
+      session: SessionInterface,
+    ) => Promise<boolean>,
+    readonly loadSessionCallback: (
       id: string,
     ) => Promise<SessionInterface | {[key: string]: unknown} | undefined>,
-    readonly deleteCallback: (id: string) => Promise<boolean>,
+    readonly deleteSessionCallback: (id: string) => Promise<boolean>,
+    readonly deleteSessionsCallback?: (
+      id: string[],
+    ) => Promise<boolean> | undefined,
```
paulomarg, line 17: I think we can _not_ return `undefined` here - if the function is defined, we should just force the response type to be a `Promise<boolean>`.

```suggestion
    ) => Promise<boolean>,
```

src/auth/session/storage/custom.ts
```diff
@@ -70,11 +82,58 @@ export class CustomSessionStorage implements SessionStorage {
...
       throw new ShopifyErrors.SessionStorageError(
         `CustomSessionStorage failed to delete a session. Error Details: ${error}`,
       );
     }
   }
+
+  public async deleteSessions(ids: string[]): Promise<boolean> {
+    if (this.deleteSessionsCallback) {
+      try {
+        return (await this.deleteSessionsCallback(ids)) as boolean;
```
paulomarg, line 96: If the function doesn't return `undefined`, we probably won't need to cast this.

src/auth/session/storage/custom.ts
```diff
@@ -70,11 +82,58 @@ export class CustomSessionStorage implements SessionStorage {
...
+      } catch (error) {
+        throw new ShopifyErrors.SessionStorageError(
+          `CustomSessionStorage failed to delete array of sessions. Error Details: ${error}`,
+        );
+      }
+    } else {
+      throw new ShopifyErrors.SessionStorageError(
```
paulomarg, line 103: I wonder if throwing an error makes sense here, because we're essentially making the optional functions mandatory if we blow up when they're called. Would it make sense to just log something here instead?

src/auth/session/storage/custom.ts
```diff
@@ -70,11 +82,58 @@ export class CustomSessionStorage implements SessionStorage {
...
+        const sessions: SessionInterface[] = [];
+
+        results.forEach((element) => {
+          sessions.push(element as SessionInterface);
+        });
+        return sessions;
+      }
+      return undefined;
+    } else {
+      throw new ShopifyErrors.SessionStorageError(
```
paulomarg, line 134: Same error comment here.

src/auth/session/storage/memory.ts
```diff
@@ -19,4 +19,18 @@ export class MemorySessionStorage implements SessionStorage {
...
+    return true;
+  }
+
+  public async findSessionsByShop(
+    shop: string,
+  ): Promise<SessionInterface[] | {[key: string]: unknown}[] | undefined> {
+    const results = Object.values(this.sessions).filter(
+      (session) => session.shop === shop,
+    );
+    return results.length === 0 ? undefined : results;
```
paulomarg, line 34: The `undefined` return here I think makes it a bit awkward - could we just always return `results`, even if it's empty?

src/auth/session/storage/mongodb.ts
```diff
@@ -75,6 +75,29 @@ export class MongoDBSessionStorage implements SessionStorage {
     return true;
   }
 
+  public async deleteSessions(ids: string[]): Promise<boolean> {
+    await this.ready;
+    for (const id of ids) {
+      await this.collection.deleteOne({id});
+    }
```
paulomarg, line 82: Instead of looping over the entries, could we just use [`deleteMany`](https://www.mongodb.com/docs/v4.4/reference/method/db.collection.deleteMany/?_ga=2.7681010.1394317485.1657638452-2132842715.1657638452#db.collection.deletemany--)?

src/auth/session/storage/mongodb.ts
```diff
@@ -75,6 +75,29 @@ export class MongoDBSessionStorage implements SessionStorage {
...
+    return true;
+  }
+
+  public async findSessionsByShop(
+    shop: string,
+  ): Promise<SessionInterface[] | {[key: string]: unknown}[] | undefined> {
+    await this.ready;
+
+    const rawResults = await this.collection.find().toArray();
```
paulomarg, line 91: I'm not super familiar with the `find` method, but would it be possible to filter by the shop we're looking for at the query level, rather than filtering out the results? Seems like that would be more efficient.

src/auth/session/storage/mysql.ts
```diff
@@ -82,6 +82,34 @@ export class MySQLSessionStorage implements SessionStorage {
...
+  public async findSessionsByShop(
+    shop: string,
+  ): Promise<SessionInterface[] | {[key: string]: unknown}[] | undefined> {
+    await this.ready;
+    const query = `
+      SELECT * FROM ${this.options.sessionTableName}
+      WHERE shop = ?;
+    `;
+    const [rows] = await this.query(query, [shop]);
+    if (!Array.isArray(rows) || rows?.length === 0) return undefined;
```
paulomarg, line 104: Comment on `return undefined` here again.

src/auth/session/storage/mysql.ts
```diff
@@ -82,6 +82,34 @@ export class MySQLSessionStorage implements SessionStorage {
...
+
+    const results: SessionInterface[] = [];
+    for (const row of rows) {
```
paulomarg, line 107: I think we can simplify this a bit by using `map` instead of a loop. I'm assuming we can probably do that for all SQL-based classes?

src/auth/session/storage/redis.ts
```diff
@@ -71,6 +71,32 @@ export class RedisSessionStorage implements SessionStorage {
     return true;
   }
 
+  public async deleteSessions(ids: string[]): Promise<boolean> {
+    await this.ready;
+    for (const id of ids) {
+      await this.deleteSession(id);
+    }
```
paulomarg, line 78: I _think_ the redis package allows us to pass in an array of ids to delete. Don't quote me on that, though.

src/auth/session/storage/redis.ts
```diff
@@ -71,6 +71,32 @@ export class RedisSessionStorage implements SessionStorage {
...
+    return true;
+  }
+
+  public async findSessionsByShop(
+    shop: string,
+  ): Promise<SessionInterface[] | {[key: string]: unknown}[] | undefined> {
+    await this.ready;
+
+    const keys = await this.client.keys('*');
```
paulomarg, line 87: I wonder if it's worth storing the ids for specific shops under a single key, so that we don't have to iterate over every entry to find the ids for a shop. Quite a big change, but I think it's worth it.

paulomarg (commented):
Looking good, I had some thoughts on how to make a couple of these more efficient, but overall I think this makes sense!

mkevinosullivan:
> This LGTM.
> 
> One question: What is the use-case for `findSessionsByShop`? Do we need to worry about the returned array being _uber_ large for popular apps?

It's "give me the sessions for this one shop", so, with the additional change of using offline sessions, it realistically should be just one entry.

src/auth/session/storage/custom.ts
```diff
@@ -5,20 +5,32 @@ import * as ShopifyErrors from '../../../error';
...
+    readonly storeSessionCallback: (
+      session: SessionInterface,
+    ) => Promise<boolean>,
+    readonly loadSessionCallback: (
       id: string,
     ) => Promise<SessionInterface | {[key: string]: unknown} | undefined>,
-    readonly deleteCallback: (id: string) => Promise<boolean>,
+    readonly deleteSessionCallback: (id: string) => Promise<boolean>,
+    readonly deleteSessionsCallback?: (
+      id: string[],
```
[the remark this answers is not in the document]
mkevinosullivan: 🤦🏻

src/auth/session/storage/custom.ts
mkevinosullivan: 👍🏻

src/auth/session/storage/custom.ts
mkevinosullivan: 👍🏻

src/auth/session/storage/custom.ts
mkevinosullivan: K, changing to `console.warn`

src/auth/session/storage/mysql.ts
mkevinosullivan: 👍🏻

src/auth/session/storage/redis.ts
mkevinosullivan: K, that would be better (anything would be better than this current search implementation!) but since it's such a big change to the underlying data structure, I'll leave it for its own PR.

mkevinosullivan:
API library component of https://github.com/Shopify/first-party-library-planning/issues/384

src/auth/session/storage/custom.ts
```diff
@@ -5,20 +5,28 @@ import * as ShopifyErrors from '../../../error';
 
 export class CustomSessionStorage implements SessionStorage {
   constructor(
-    readonly storeCallback: (session: SessionInterface) => Promise<boolean>,
-    readonly loadCallback: (
+    readonly storeSessionCallback: (
```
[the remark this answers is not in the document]
mkevinosullivan: Yeah, it's a fair point about it potentially be breaking.  In theory, they _shouldn't_ call `myStorage.storeCallback`, since the goal is to have something that responds to `storeSession` by the library when needed, but ... 

I thought it would be nice to be consistent but I'll change them back to avoid problems.

PR #419, opened 2022-07-12
Add a new utility for constructing the host URL for the app.

byrichardpowell:
Soon we will be recommending that apps favor server redirects within the OAuth process, rather than client side redirects using AppBridge.  To make this a little easier, we want to control that code that constructs the URL for the app.  That way it's easier for developers, and if that URL changes we can update the logic with a version bump.

Add a new utility called `getHostAppUrl`.  This utility constructs the host URL for an app.  

This utility will be used to get the URL to redirect to when the store does not have a token for the app.

*  [Example here](https://github.com/Shopify/shopify-app-template-node/compare/cli_three...load-inside-admin?expand=1#diff-1107ca59bb2d9a8cd268c58eb15e6d426d68d37371f73abbb8cc47a6fd75298aR172-R177)
* [Another example here](https://github.com/Shopify/shopify-app-template-node/compare/cli_three...load-inside-admin?expand=1#diff-339f370288e8edd7c41685f6e82fca547f5ff3931df26e042a893d8b21f9ea29R45-R50)

## Type of change

## Checklist

### DIFF
--- a/src/utils/__tests__/get-host-app-url.test.ts
+++ b/src/utils/__tests__/get-host-app-url.test.ts
@@ -1,0 +1,47 @@
+import http from 'http';
+
+import getHostAppUrl from '../get-host-app-url';
+import * as ShopifyErrors from '../../error';
+import {Context} from '../../context';
+
+describe('getHostAppUrl', () => {
+  beforeEach(() => {
+    Context.API_KEY = 'my-api-key';
+  });
+
+  test('thows an error when no request is passed', () => {
+    // @ts-expect-error: For JS users test it throws when no request is passed
+    expect(() => getHostAppUrl()).toThrow(
+      ShopifyErrors.MissingRequiredArgument,
+    );
+  });
+
+  test('thows an error when the request has no URL', () => {
+    const req = {
+      url: undefined,
+    } as http.IncomingMessage;
+
+    expect(() => getHostAppUrl(req)).toThrow(ShopifyErrors.InvalidRequestError);
+  });
+
+  test('thows an error when the request has no host query param', () => {
+    const req = {
+      url: 'www.example.com',
+    } as http.IncomingMessage;
+
+    expect(() => getHostAppUrl(req)).toThrow(ShopifyErrors.InvalidRequestError);
+  });
+
+  test('returns the host app url', () => {
+    const host = 'my-store.shopify.com';
+    const base64Host = Buffer.from(host, 'utf-8').toString('base64');
+
+    const req = {
+      url: `www.example.com?host=${base64Host}`,
+    } as http.IncomingMessage;
+
+    expect(getHostAppUrl(req)).toBe(
+      'https://my-store.shopify.com/apps/my-api-key',
+    );
+  });
+});
--- a/src/utils/get-host-app-url.ts
+++ b/src/utils/get-host-app-url.ts
@@ -1,0 +1,36 @@
+import http from 'http';
+import url from 'url';
+
+import * as ShopifyErrors from '../error';
+import {Context} from '../context';
+
+/**
+ * Helper method to get the host URL for the app.
+ *
+ * @param request Current HTTP request
+ */
+export default function getHostAppUrl(request: http.IncomingMessage): string {
+  if (!request) {
+    throw new ShopifyErrors.MissingRequiredArgument(
+      'getHostAppUrl requires a request object argument',
+    );
+  }
+
+  if (!request.url) {
+    throw new ShopifyErrors.InvalidRequestError(
+      'Request does not contain a URL',
+    );
+  }
+
+  const query = url.parse(request.url, true).query;
+
+  if (typeof query.host !== 'string') {
+    throw new ShopifyErrors.InvalidRequestError(
+      'Request does not contain a host query parameter',
+    );
+  }
+
+  const host = Buffer.from(query.host, 'base64').toString();
+
+  return `https://${host}/apps/${Context.API_KEY}`;
+}
--- a/src/error.ts
+++ b/src/error.ts
@@ -75,3 +75,5 @@ export class SessionStorageError extends ShopifyError {}
 
 export class MissingRequiredArgument extends ShopifyError {}
 export class UnsupportedClientType extends ShopifyError {}
+
+export class InvalidRequestError extends ShopifyError {}
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -11,6 +11,7 @@ import validateHmac from './hmac-validator';
 import validateShop from './shop-validator';
 import versionCompatible from './version-compatible';
 import withSession from './with-session';
+import getHostAppUrl from './get-host-app-url';
 
 const ShopifyUtils = {
   decodeSessionToken,
@@ -26,6 +27,7 @@ const ShopifyUtils = {
   validateShop,
   versionCompatible,
   withSession,
+  getHostAppUrl,
 };
 
 export default ShopifyUtils;


byrichardpowell:
@paulomarg @vividviolet this is now ready for review.  I'm going to test the code manually in the app template.

docs/usage/utils.md
```diff
@@ -0,0 +1,12 @@
...
+## Get the HOST application URL using `getHostAppUrl`
+
+If you need to redirect a request to your app inside the user's admin you can use `getHostAppUrl`.
+
+```ts
+const redirectURL = getHostAppUrl(request);
+res.redirect(redirectURL);
+```
+
+This utility relies on the host query param being a Base 64 encoded string. All requests from Shopify should include this param in the correct format.
```
byrichardpowell, line 12: @paulomarg  I wasn't really sure what to say here since the utility is so small.

src/utils/__tests__/get-host-app-url.test.ts
```diff
@@ -0,0 +1,47 @@
...
+import getHostAppUrl from '../get-host-app-url';
+import * as ShopifyErrors from '../../error';
+import {Context} from '../../context';
+
+describe('getHostAppUrl', () => {
+  beforeEach(() => {
+    Context.API_KEY = 'my-api-key';
+  });
+
+  test('thows an error when no request is passed', () => {
```
mkevinosullivan, line 12:
```suggestion
  test('throws an error when no request is passed', () => {
```

src/utils/__tests__/get-host-app-url.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    // @ts-expect-error: For JS users test it throws when no request is passed
+    expect(() => getHostAppUrl()).toThrow(
+      ShopifyErrors.MissingRequiredArgument,
+    );
+  });
+
+  test('thows an error when the request has no URL', () => {
```
mkevinosullivan, line 19:
```suggestion
  test('throws an error when the request has no URL', () => {
```

src/utils/__tests__/get-host-app-url.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    const req = {
+      url: undefined,
+    } as http.IncomingMessage;
+
+    expect(() => getHostAppUrl(req)).toThrow(ShopifyErrors.InvalidRequestError);
+  });
+
+  test('thows an error when the request has no host query param', () => {
```
mkevinosullivan, line 27:
```suggestion
  test('throws an error when the request has no host query param', () => {
```

src/utils/get-host-app-url.ts
```diff
@@ -0,0 +1,36 @@
...
+    );
+  }
+
+  if (!request.url) {
+    throw new ShopifyErrors.InvalidRequestError(
+      'Request does not contain a URL',
+    );
+  }
+
+  const query = url.parse(request.url, true).query;
```
mkevinosullivan, line 25: `url.parse` is showing up as *legacy* though not deprecated (because so many modules rely on it) but it has security concerns: https://nodejs.org/api/url.html#urlparseurlstring-parsequerystring-slashesdenotehost

Wondering if this should be 
```suggestion
  const host = new URL(request.url).searchParams.get('host');
```
(which is terribly clunky 😞)

src/utils/__tests__/get-host-app-url.test.ts
byrichardpowell: Oops, thanks Kevin 👍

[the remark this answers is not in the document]
byrichardpowell: Oops, thanks Trish 👍

src/utils/get-host-app-url.ts
```diff
@@ -0,0 +1,36 @@
...
+ * @param request Current HTTP request
+ */
+export default function getHostAppUrl(request: http.IncomingMessage): string {
+  if (!request) {
+    throw new ShopifyErrors.MissingRequiredArgument(
+      'getHostAppUrl requires a request object argument',
+    );
+  }
+
+  if (!request.url) {
```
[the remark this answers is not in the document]
byrichardpowell: Yeah, this really surprised me, but according to the type definitions it can be undefined:

Screen Shot 2022-07-12 at 5 05 31 PM

Unless I've mis-understood something?

src/utils/get-host-app-url.ts
```diff
@@ -0,0 +1,36 @@
...
+
+import * as ShopifyErrors from '../error';
+import {Context} from '../context';
+
+/**
+ * Helper method to get the host URL for the app.
+ *
+ * @param request Current HTTP request
+ */
+export default function getHostAppUrl(request: http.IncomingMessage): string {
```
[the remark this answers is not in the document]
byrichardpowell: Yeah, true.  

I copied the format we've used elsewhere in the library.  E.g:

```
export default async function graphqlProxy(
  userReq: http.IncomingMessage,
  userRes: http.ServerResponse,
): Promise<RequestReturn> {
```

and:

```
  async beginAuth(
    request: http.IncomingMessage,
    response: http.ServerResponse,
    shop: string,
    redirectPath: string,
    isOnline = true,
  ): Promise<string> {
```

This also applies to `deleteCurrentSession`, `getEmbeddedAppUrl`, `loadCurrentSession` and a few other places.

Is it be better to be consistent here?  Thoughts @paulomarg @henrytao-me ?

src/utils/__tests__/get-host-app-url.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    const req = {
+      url: 'www.example.com',
+    } as http.IncomingMessage;
+
+    expect(() => getHostAppUrl(req)).toThrow(ShopifyErrors.InvalidRequestError);
+  });
+
+  test('returns the host app url', () => {
+    const host = 'my-store.shopify.com';
```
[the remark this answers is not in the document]
vividviolet: That's what the shape of the host param already look like after you decode it - it includes `/admin` in the path

docs/usage/utils.md
vividviolet: Yes, we can say that using `host` will ensure the redirect url is properly constructed and brings the merchant to the right place and is more reliably than using `shop`

byrichardpowell: Updated 👍

src/utils/get-host-app-url.ts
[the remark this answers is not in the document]
byrichardpowell: Perfect, thanks 👍

src/utils/get-host-app-url.ts
byrichardpowell: Thanks for spotting that.  Ive updated this to use `URL`.

henrytao-me (approved):
LGTM 👍

src/utils/get-host-app-url.ts
mkevinosullivan: Is it possible that `request.url` will not be absolute?  If it will always be absolute, then the first line could be simplified to

```typescript
const url = new URL(request.url);
```

Otherwise, this seems good to me!

mkevinosullivan: If we need to keep the `base` parameter for the `new URL()` call, should it be `https` instead of `http`?

src/utils/get-host-app-url.ts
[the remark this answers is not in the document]
vividviolet: I thought url was always a string according to this https://nodejs.org/api/http.html#messageurl

src/utils/get-host-app-url.ts
byrichardpowell: In my testing the express template passed a relative URL.

I'll update it to use https though.  I don't think it would matter as we don't use that portion of the URL, but https is a better default.

PR #426, opened 2022-07-20
Include query body in API deprecation logs

paulomarg:

Related to https://github.com/Shopify/cli/issues/156

Currently, we're logging deprecation notices, but for the GraphQL API, we don't get a lot of details - since every request lands in the same endpoint, there's no way to tell them apart.

Adding the first 100 characters from the body (if one exists) to help indicate which query is triggering the logs, to make it easier for developers to figure out where they're hitting deprecated fields.

```
API Deprecation Notice: {
  message: 'https://shopify.dev/api/usage/versioning#deprecation-practices',
  path: 'https://shop.myshopify.com/admin/api/2022-07/graphql.json',
  body: 'query products {\\n        products(first: 2) {\\n          nodes {\\n            title\\n            bodyHt...'
}
```

## Type of change

## Checklist

### DIFF
--- a/src/clients/http_client/http_client.ts
+++ b/src/clients/http_client/http_client.ts
@@ -20,6 +20,12 @@ import {
   RequestReturn,
 } from './types';
 
+interface DeprecationInterface {
+  message: string | null;
+  path: string;
+  body?: string;
+}
+
 export class HttpClient {
   // 1 second
   static readonly RETRY_WAIT_TIME = 1000;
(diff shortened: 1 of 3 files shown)


src/clients/http_client/http_client.ts
```diff
@@ -182,11 +188,25 @@ export class HttpClient {
...
           response.headers.has('X-Shopify-API-Deprecated-Reason')
         ) {
-          const deprecation = {
+          const deprecation: DeprecationInterface = {
             message: response.headers.get('X-Shopify-API-Deprecated-Reason'),
             path: url,
           };
 
+          if (options.body) {
+            if (typeof options.body === 'string') {
```
mkevinosullivan, line 197: Is there really a need for this since https://github.com/Shopify/shopify-api-node/pull/426/files#diff-1d5aeaa437fd851787c18a3ca142885c050d28a290937f6e4d5ca8afea9d1ccfR99-R112 essentially guarantees it will be a string?

paulomarg: I had to do this because of the typing of `options.body`, which can be a few different things.

PR #431, opened 2022-07-22
Add handling of `errors` field in GQL response

mkevinosullivan:

GraphQL will return 200 OK even if there was an issue with the GraphQL query.

API lib part of https://github.com/Shopify/first-party-library-planning/issues/385

This change adds a check to see if the response contains an `errors` field and will throw `GraphqlQueryError` if it does. It also includes a stringified version of the response body in the error.

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 ## Unreleased
 
 - Allow passing in options for the Redis client used by the session storage strategy [#430](https://github.com/Shopify/shopify-api-node/pull/430)
-- If a response from a GraphQL query contains an `errors` attribute, `GraphqlClient` will now throw a `GraphqlQueryError`. The caller can check the `error.response` attribute to see what was returned from the GraphQL API. [#431](https://github.com/Shopify/shopify-api-node/pull/431)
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+- If a response from a GraphQL query contains an `errors` attribute, `GraphqlClient` will now throw a `GraphqlQueryError`. The caller can check the `error.response` attribute to see what was returned from the GraphQL API. [#431](https://github.com/Shopify/shopify-api-node/pull/431)
```
paulomarg, line 10: We should add a breaking marker here to make it clear folks will need to update it.

src/webhooks/registry.ts
```diff
@@ -261,20 +261,31 @@ const WebhooksRegistry: RegistryInterface = {
...
+        registerReturn[topic] = {
+          success: false,
+          result: error.response,
+        };
+      } else {
+        registerReturn[topic] = {
+          success: false,
+          result: {},
+        };
+      }
```
paulomarg, line 281: Nit: we could simplify this a bit so that we only need to write the response object once:
```suggestion
      const result = (error instanceof ShopifyErrors.GraphqlQueryError) ? error.response : {};
      registerReturn[topic] = { success: false, result };
```

PR #447, opened 2022-08-05
Dynamically type request responses

paulomarg:

Currently, when getting data from either the GraphQL or REST clients, the caller needs to manually cast the returned type which makes it more difficult than it needs to be to type the responses properly.

Adding the option of typing the bodies of request responses when calling the appropriate methods.

## Type of change

## Checklist

### DIFF
--- a/src/clients/graphql/__tests__/graphql_client.test.ts
+++ b/src/clients/graphql/__tests__/graphql_client.test.ts
@@ -203,28 +203,6 @@ describe('GraphQL client', () => {
       data: JSON.stringify(query),
     }).toMatchMadeHttpRequest();
   });
-
-  it('properly types the response', async () => {
-    const client: GraphqlClient = new GraphqlClient(DOMAIN, 'bork');
-
-    const typedResponse = {
-      value1: 1,
-      value2: false,
-    };
-
-    fetchMock.mockResponseOnce(JSON.stringify(typedResponse));
-
-    const response = await client.query<typeof typedResponse>({data: QUERY});
-
-    expect({
-      method: 'POST',
-      domain: DOMAIN,
-      path: '/admin/api/unstable/graphql.json',
-      data: QUERY,
-    }).toMatchMadeHttpRequest();
-
-    expect(typeof response).toBe(typeof typedResponse);
-  });
 });
 
 function buildExpectedResponse(obj: unknown) {
--- a/src/clients/graphql/__tests__/graphql_client.test.ts
+++ b/src/clients/graphql/__tests__/graphql_client.test.ts
@@ -203,6 +203,28 @@ describe('GraphQL client', () => {
       data: JSON.stringify(query),
     }).toMatchMadeHttpRequest();
   });
+
+  it('properly types the response', async () => {
+    const client: GraphqlClient = new GraphqlClient(DOMAIN, 'bork');
+
+    const typedResponse = {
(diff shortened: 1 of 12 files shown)


src/clients/graphql/__tests__/graphql_client.test.ts
```diff
@@ -203,6 +203,28 @@ describe('GraphQL client', () => {
       data: JSON.stringify(query),
     }).toMatchMadeHttpRequest();
   });
+
+  it('properly types the response', async () => {
```
paulomarg, line 207: Does this test make sense - i.e. is it actually going to test anything sane at run time?

JaKXz: Unfortunately I don't think so 😅 but if tsc is set to include test files it will type check the file in the compile step.

paulomarg: We have usage of these types here and there so the type sanity will be checked eventually. I'll remove the pointless test :)

src/clients/rest/rest_client.ts
```diff
@@ -76,7 +78,7 @@ class RestClient extends HttpClient {
       ret.pageInfo = pageInfo;
     }
 
-    return ret;
+    return ret as unknown as RestRequestReturn<T>;
```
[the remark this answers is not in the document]
paulomarg: We do not! I'm cleaning this method up a little!

src/clients/graphql/graphql_client.ts
```diff
@@ -26,7 +26,7 @@ export class GraphqlClient {
     this.client = new HttpClient(this.domain);
   }
 
-  async query(params: GraphqlParams): Promise<RequestReturn> {
+  async query<T = unknown>(params: GraphqlParams): Promise<RequestReturn<T>> {
```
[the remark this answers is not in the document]
paulomarg: We can look into improving the default type here - right now this will map to `body: unknown` by default, so we could refactor that default later on to be more representative of an actual response.

teddyhwang (approved):

src/clients/rest/rest_client.ts
```diff
@@ -22,15 +22,17 @@ class RestClient extends HttpClient {
...
+  ): Promise<RestRequestReturn<T>> {
     params.extraHeaders = {
       [ShopifyHeader.AccessToken]: Context.IS_PRIVATE_APP
         ? Context.API_SECRET_KEY
         : (this.accessToken as string),
       ...params.extraHeaders,
     };
 
-    const ret = (await super.request(params)) as RestRequestReturn;
+    const ret = (await super.request<T>(params)) as RestRequestReturn<T>;
```
JaKXz, line 35: Part of my previous suggestion `HttpClient#request` should also accept a type param no? and then we can just pass
```suggestion
    const ret = await super.request<RestRequestReturn>(params);
```

like this? Maybe I'm misreading something?

paulomarg: So, I thought about this but I _think_ `RestRequestReturn` is at the same level as `RequestReturn`, so I think typing the `request` call would end up returning `RequestReturn<RestRequestReturn<T>>`, which isn't the goal here.

In this case I think we actually need the cast because `RestRequestReturn` is a superset of `RequestReturn` rather than a component.

PR #451, opened 2022-08-15
Reorganize tests into jest projects

paulomarg:
While adding the tests for https://github.com/Shopify/shopify-api-node/pull/450, it was becoming too difficult to add ignore patterns for all of the folders with separate test suites in the main suite command.

This PR makes that easier by creating separate jest projects, which allows us to fork the configs into different bundles so that we can more freely tweak them for each suite.


mkevinosullivan (approved):
That cleans it up/structures it quite nicely 🔥

PR #455, opened 2022-08-18
Running the adapter tests for CF workers using miniflare

paulomarg:

We have some generic tests for the adapter implementations of the `runtime/http` and `runtime/crypto` features, which were previously only run for the node adapter. That made it impossible to find issues with the worker implementation of those features.

Adding a new jest project for CF worker tests, and running that same suite of tests for that adapter using it.


src/runtime/crypto/__tests__/produce-node-hmac.mjs
```diff
@@ -0,0 +1,33 @@
+import crypto from 'crypto';
+
+function random({min = 0, max = 1, floor = true} = {}) {
+  let value = Math.random() * (max - min) + min;
+  if (floor) {
+    value = Math.floor(value);
+  }
+  return value;
+}
```
mkevinosullivan, line 9: Since this is duplicated in `produce-node-base64.mjs`, opportunity to put in a shared file and imported into each script?

mkevinosullivan (approved):

PR #463, opened 2022-08-22
Refactoring Context into config

paulomarg:

The `Shopify.Context` pattern is a bit clunky from a "typical" JS app's perspective. This PR aims to improve that by breaking it up into a global `config` object to read values, and a `setConfig` function to write them.

As a way of validating the migration doc, this is what we're changing. Please let me know if it's unclear:

image


docs/migrating_to_isomorphic.md
```diff
@@ -14,3 +14,39 @@ This automatically sets the library up to run on the Node.js runtime.
...
+Shopify.Context.initialize({ ... });
+```
+
+</td>
+<td>
+
+```ts
+import {setConfig, config} from '@shopify/shopify-api';
+setConfig({ SHOPIFY_API_KEY: '...', ... });
+console.log(config.SHOPIFY_API_KEY);
```
paulomarg, line 47: Which one works better as an example?

```suggestion
import * as Shopify from '@shopify/shopify-api';
Shopify.setConfig({ SHOPIFY_API_KEY: '...', ... });
console.log(Shopify.config.SHOPIFY_API_KEY);
```

docs/issues.md
```diff
@@ -4,11 +4,11 @@ By following this guide, you will have a fully functional Shopify app. However,
 
 ## Notes on session handling
 
-Before you start writing your application, please note that the Shopify library stores some information for OAuth in sessions. Since each application may choose a different strategy to store information, the library cannot dictate any specific storage strategy. By default, `Shopify.Context` is initialized with `SQLiteSessionStorage`, which will enable you to start developing your app by storing sessions using the file-based SQLite package. You can quickly get your app set up using it, but please keep the following in mind.
+Before you start writing your application, please note that the Shopify library stores some information for OAuth in sessions. Since each application may choose a different strategy to store information, the library cannot dictate any specific storage strategy. By default, `config` is initialized with `SQLiteSessionStorage`, which will enable you to start developing your app by storing sessions using the file-based SQLite package. You can quickly get your app set up using it, but please keep the following in mind.
```
paulomarg, line 7: For now, this is temporarily using `Memory`, but we might want to make it configurable per runtime adapter, since there will likely not be one one-size-fits-all solution.

src/config.ts
```diff
@@ -0,0 +1,105 @@
+import {ShopifyError, UninitializedConfigError, PrivateAppError} from './error';
```
paulomarg, line 1: This is probably the most important place to check (and its test file). Anything else is just search / replacing for the name change.

docs/issues.md
paulomarg: Setting up a separate PR to tackle that as this one is too big with all the search-replace going on.

src/config.ts
```diff
@@ -0,0 +1,105 @@
...
+    missing.push('hostName');
+  }
+
+  if (missing.length) {
+    throw new ShopifyError(
+      `Cannot initialize Shopify API Library. Missing values for: ${missing.join(
+        ', ',
+      )}`,
+    );
+  }
```
mkevinosullivan, line 56: By rearranging this a bit and adding a `.length` property to `AuthScopes` (in case `param.scopes` is already that type on entry), this could be tidied up to be something like the following, which would allow relatively easy addition of other mandatory keys in future.
```suggestion
  // Make sure that the essential params actually have content in them
  const mandatory: string[]= ['apiKey', 'apiSecretKey', 'scopes', 'hostName'];
  const missing: string[] = [];
  keys.forEach((key) => { if (!params[key].length) { missing.push(key); }});

  if (missing.length) {
    throw new ShopifyError(
      `Cannot initialize Shopify API Library. Missing values for: ${missing.join(
        ', ',
      )}`,
    );
  }

  if (params.scopes instanceof AuthScopes) {
    config.scopes = params.scopes;
  } else {
    config.scopes = new AuthScopes(params.scopes);
  }
```

src/config.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  if (params.customShopDomains) {
+    config.customShopDomains = params.customShopDomains;
+  }
+
+  if (params.billing) {
+    config.billing = params.billing;
+  }
+}
```
mkevinosullivan, line 91: Similarly, this could be compacted to:
```suggestion
  const optional: string[] = [
    'hostScheme',
    'userAgentPrefix',
    'logFile',
    'privateAppStorefrontAccessToken',
    'customShopDomains',
    'billing'
  ];
  
  optional.forEach((key) => { if (params[optional]) { config[key] = param[key]; }});
}
```

src/config.ts
```diff
@@ -0,0 +1,105 @@
...
+
+export function throwIfUninitializedConfig(): void {
```
mkevinosullivan, line 93: Just to simplify the wording a bit?
```suggestion
export function throwIfConfigNotSet(): void {
```

src/error.ts
```diff
@@ -12,7 +12,7 @@ export class InvalidJwtError extends ShopifyError {}
 export class MissingJwtTokenError extends ShopifyError {}
 
 export class SafeCompareError extends ShopifyError {}
-export class UninitializedContextError extends ShopifyError {}
+export class UninitializedConfigError extends ShopifyError {}
```
mkevinosullivan, line 15:
```suggestion
export class ConfigNotSetError extends ShopifyError {}
```

docs/usage/webhooks.md
```diff
@@ -202,8 +202,8 @@ app.post('/webhooks', async (req, res) => {
 
 ### Note regarding use of body parsers
 
-Please note that the use of body parsing middleware must occur **after** webhook processing.  `Shopify.Webhooks.Registry.process()` reads in the request body directly, therefore, if a body parsing middleware is used beforehand, `process` thinks the request body is empty and will return a `bad request` message back to Shopify for the webhook and raise an error.
+Please note that the use of body parsing middleware must occur **after** webhook processing. `Shopify.Webhooks.Registry.process()` reads in the request body directly, therefore, if a body parsing middleware is used beforehand, `process` thinks the request body is empty and will return a `bad request` message back to Shopify for the webhook and raise an error.
```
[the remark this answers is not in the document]
paulomarg: The idea is that we can revisit the exported interfaces of each component (and their docs) as we refactor them into isomorphic ones.

src/config.ts
```diff
@@ -0,0 +1,105 @@
...
+  scopes: AuthScopes;
+}
+
+export const config: ConfigInterface = {
+  apiKey: '',
+  apiSecretKey: '',
+  scopes: new AuthScopes([]),
+  hostName: '',
+  hostScheme: 'https',
+  apiVersion: ApiVersion.Unstable,
```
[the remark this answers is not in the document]
paulomarg: Good callout!

mkevinosullivan (approved):

mkevinosullivan (approved):

PR #472, opened 2022-08-29
Add utils to main object

paulomarg:

Many of the utils functions relied on the global config being available. With it gone, those utils would break, so we'll need some way to call them while including the config

For now, we're creating a "global" utils object in the exported library object, but we'll later refactor these individually so they're placed in a better place semantically.

### DIFF
--- a/src/utils/hmac-validator.ts
+++ b/src/utils/hmac-validator.ts
@@ -1,6 +1,7 @@
 import crypto from 'crypto';
 import querystring from 'querystring';
 
+import {ConfigInterface} from '../base-types';
 import {AuthQuery} from '../auth/oauth/types';
 import * as ShopifyErrors from '../error';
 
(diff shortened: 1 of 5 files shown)


src/utils/hmac-validator.ts
```diff
@@ -20,13 +19,10 @@ export function stringifyQuery(query: AuthQuery): string {
...
 
-export function generateLocalHmac({
-  code,
-  timestamp,
-  state,
-  shop,
-  host,
-}: AuthQuery): string {
+export function generateLocalHmac(
+  apiSecretKey: string,
```
mkevinosullivan, line 23: Putting a note here that this is changed in another branch, so these values can be obtained from config, when `generateLocalHmac` is attached using a `create__(config)` type function.

src/utils/hmac-validator.ts
```diff
@@ -35,24 +31,19 @@ export function generateLocalHmac({
...
     .digest('hex');
 }
 
-/**
- * Uses the received query to validate the contained hmac value against the rest of the query content.
- *
- * @param query HTTP Request Query, containing the information to be validated.
- */
-export default function validateHmac(query: AuthQuery): boolean {
+export function validateHmac(apiSecretKey: string, query: AuthQuery): boolean {
```
mkevinosullivan, line 39: Putting a note here that this is changed in another branch, so these values can be obtained from config, when `validateHmac` is attached using a `create__(config)` type function.

src/utils/index.ts
```diff
@@ -1,34 +1,40 @@
...
+import {createGetEmbeddedAppUrl} from './get-embedded-app-url';
 
-export default ShopifyUtils;
+export function shopifyUtils(config: ConfigInterface) {
+  return {
+    decodeSessionToken,
+    // deleteCurrentSession: createDeleteCurrentSession(config),
+    // deleteOfflineSession: createDeleteOfflineSession(config),
+    // loadCurrentSession: createLoadCurrentSession(config),
+    // loadOfflineSession: createLoadOfflineSession(config),
```
mkevinosullivan, line 29: To be added later?

src/utils/with-session.ts
```diff
@@ -1,66 +1,68 @@
 import * as ShopifyErrors from '../error';
-import {Session} from '../auth/session';
+import {SessionInterface} from '../auth/session/types';
 import {GraphqlClient} from '../clients/graphql';
 import {RestClient} from '../clients/rest';
```
mkevinosullivan, line 4: Will need to be refactored after rebase, given these types no longer exist

mkevinosullivan (approved):

src/utils/index.ts
paulomarg: Yes, these will be fixed as part of oauth, there are a few interdependencies there so I felt it was better to tackle them all at once.

PR #477, opened 2022-09-02
Use mock adapter for tests

paulomarg:

We want to test the library generically rather than default to the node adapter and create mock `http.IncomingMessage` and `http.ServerResponse` objects for testing.

Creating a new mock adapter which simply passes the normalized objects through to make it easier to test the library in a more generic way.


src/utils/get-embedded-app-url.ts
```diff
@@ -1,12 +1,11 @@
...
-
 import * as ShopifyErrors from '../error';
 import {ConfigInterface} from '../base-types';
+import {NormalizedRequest} from '../runtime/http';
 
 import {createSanitizeHost} from './shop-validator';
 
 export function createGetEmbeddedAppUrl(config: ConfigInterface) {
-  return (request: http.IncomingMessage): string => {
+  return (request: NormalizedRequest): string => {
```
mkevinosullivan, line 8: As per Slack chat, this should be hidden away from the dev ... so, using `AdapterArgs` as the param type, the dev need only pass in (`{rawRequest: req}`) for whatever platform they're using and we can hide the conversion internally.

mkevinosullivan (changes requested):

PR #484, opened 2022-09-07
Refactor clients and tests to use isomorphic mock

mkevinosullivan:

The `HttpClient` relied on `node-fetch` for `fetch()`, which means it wouldn't work in a CF Worker environment.

`HttpClient` is refactored to use `abstractFetch` and therefore should be independent of the platform it's on (i.e., uses isomorphic framework).

Also included in this is the removal of `node-fetch` and `jest-fetch-mock` from the mock adapter.

### DIFF
--- a/src/__tests__/test-helper.ts
+++ b/src/__tests__/test-helper.ts
@@ -6,7 +6,7 @@ import {MemorySessionStorage} from '../session/storage/memory';
 import {JwtPayload} from '../utils/types';
 import {getHMACKey} from '../utils/get-hmac-key';
 import {mockTestRequests} from '../adapters/mock/mock_test_requests';
-import {NormalizedResponse} from '../runtime/http';
+import {canonicalizeHeaders, NormalizedResponse} from '../runtime/http';
 import {RequestReturn} from '../clients/http_client/types';
 
 declare global {
@@ -69,10 +69,9 @@ export function queueMockResponse(
   partial: Partial<NormalizedResponse> = {},
 ) {
   mockTestRequests.queueResponse({
-    statusCode: 200,
-    statusText: 'OK',
-    headers: {},
-    ...partial,
+    statusCode: partial.statusCode ?? 200,
+    statusText: partial.statusText ?? 'OK',
+    headers: canonicalizeHeaders(partial.headers ?? {}),
     body,
   });
 }
--- a/src/__tests__/test-helper.ts
+++ b/src/__tests__/test-helper.ts
@@ -5,6 +5,19 @@ import {ConfigParams, LATEST_API_VERSION, Shopify} from '../base-types';
 import {MemorySessionStorage} from '../session/storage/memory';
 import {JwtPayload} from '../utils/types';
 import {getHMACKey} from '../utils/get-hmac-key';
+import {queueResponse} from '../adapters/mock/adapter';
+import {NormalizedResponse} from '../runtime/http';
+import {RequestReturn} from '../clients/http_client/types';
+
+declare global {
+  // eslint-disable-next-line @typescript-eslint/no-namespace
+  namespace jest {
+    interface Matchers<R> {
+      toBeWithinSecondsOf(compareDate: number, seconds: number): R;
+      toMatchMadeHttpRequest(): R;
+    }
+  }
+}
 
 // eslint-disable-next-line import/no-mutable-exports
 let shopify: Shopify;
@@ -36,4 +49,38 @@ export async function signJWT(
   return new jose.SignJWT(payload as any)
     .setProtectedHeader({alg: 'HS256'})
     .sign(getHMACKey(secret));
+}
+
+export function buildMockResponse(obj: unknown): string {
+  return JSON.stringify(obj);
+}
+
+export function buildExpectedResponse(obj: unknown): RequestReturn {
+  const expectedResponse: RequestReturn = {
+    body: obj,
+    headers: expect.objectContaining({}),
+  };
+
+  return expect.objectContaining(expectedResponse);
+}
+
+export function queueMockResponse(
+  body: string,
+  partial: Partial<NormalizedResponse> = {},
+) {
+  queueResponse({
+    statusCode: 200,
+    statusText: 'OK',
+    headers: {},
+    ...partial,
+    body,
+  });
+}
+
+export function queueMockResponses(
+  ...responses: Parameters<typeof queueMockResponse>[]
+) {
+  for (const [body, response] of responses) {
+    queueMockResponse(body, response);
+  }
 }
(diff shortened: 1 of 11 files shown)


src/__tests__/test-helper.ts
```diff
@@ -5,6 +5,19 @@ import {ConfigParams, LATEST_API_VERSION, Shopify} from '../base-types';
...
+
+declare global {
+  // eslint-disable-next-line @typescript-eslint/no-namespace
+  namespace jest {
+    interface Matchers<R> {
+      toBeWithinSecondsOf(compareDate: number, seconds: number): R;
+      toMatchMadeHttpRequest(): R;
+    }
+  }
+}
```
mkevinosullivan, line 20: Moved here to provide Intellisense for all tests.

src/adapters/mock/adapter.ts
```diff
@@ -26,20 +23,75 @@ export async function mockConvertResponse(
...
+  }
+  return requestList.shift()!;
+}
+
+export function getMostRecentRequest(): NormalizedRequest {
+  if (requestList.length === 0) {
+    throw new Error('No requests have been made');
+  }
+  return requestList.pop()!;
+}
```
mkevinosullivan, line 43: I don't actually use these anywhere, in preference for just `findRequest()` (below).  Not sure if there's a need for them.

src/clients/http_client/__tests__/http_client.test.ts
```diff
@@ -338,8 +345,19 @@ describe('HTTP client', () => {
...
+    // eslint-disable-next-line no-warning-comments
+    // TODO: fix this test
+    // class MyError extends Error {
+    //   constructor(...args: any) {
+    //     super(...args);
+    //     Object.setPrototypeOf(this, new.target.prototype);
+    //   }
+    // }
+    // queueError(new MyError());
+    // await testErrorResponse(null, MyError, false);
```
mkevinosullivan, line 360: This is copied from Surma's isomorphic test suite ... can't get it to work here though ... wondering if I need to?

src/clients/http_client/__tests__/http_client.test.ts
```diff
@@ -424,6 +442,14 @@ describe('HTTP client', () => {
         ),
       },
     }).toMatchMadeHttpRequest();
+  });
+
+  it('extends a User-Agent provided by config and an extra header', async () => {
```
mkevinosullivan, line 447: Split out this test because the mock adapter is not comparing headers when searching for comparison, hence was finding the first one in the pair because all the other details are the same ... could fake it by changing the URL for each test ... 🤔 but is it a huge deal if they're separate tests?

src/clients/http_client/__tests__/http_client.test.ts
```diff
@@ -828,16 +854,16 @@ describe('HTTP client', () => {
         body: {errors: 'Error 500'},
         code: 500,
         statusText: 'Error 500',
-        headers: {'X-Text-Header': ['Error 500']},
+        headers: {'X-Text-Header': 'Error 500'},
```
mkevinosullivan, line 857: (1/3) I'm a bit concerned about removing the `[]` from the comparison to get this to pass ... but this is the format we're expecting/intending to be returned for headers?

src/clients/http_client/__tests__/http_client.test.ts
```diff
@@ -850,7 +876,7 @@ describe('HTTP client', () => {
         body: {errors: 'Error 429'},
         code: 429,
         statusText: 'Error 429',
-        headers: {'X-Text-Header': ['Error 429'], 'Retry-After': ['100']},
+        headers: {'X-Text-Header': 'Error 429', 'Retry-After': '100'},
```
mkevinosullivan, line 879: (2/3) I'm a bit concerned about removing the `[]` from the comparison to get this to pass ... but this is the format we're expecting/intending to be returned for headers?

src/clients/http_client/__tests__/http_client.test.ts
```diff
@@ -872,7 +898,7 @@ describe('HTTP client', () => {
         body: {errors: 'Error 403'},
         code: 403,
         statusText: 'Error 403',
-        headers: {'X-Text-Header': ['Error 403']},
+        headers: {'X-Text-Header': 'Error 403'},
```
mkevinosullivan, line 901: (3/3) I'm a bit concerned about removing the `[]` from the comparison to get this to pass ... but this is the format we're expecting/intending to be returned for headers?

src/__tests__/test-helper.ts
```diff
@@ -5,6 +5,19 @@ import {ConfigParams, LATEST_API_VERSION, Shopify} from '../base-types';
 import {MemorySessionStorage} from '../session/storage/memory';
 import {JwtPayload} from '../utils/types';
 import {getHMACKey} from '../utils/get-hmac-key';
+import {mockAdapter} from '../adapters/mock/adapter';
```
paulomarg, line 8: I wonder if we could name this in a way that indicates it's just used to mock requests for tests, because the adapters aren't "supposed" to export anything. Maybe `mockHttpRequests` or some such?

src/adapters/mock/adapter.ts
```diff
@@ -32,16 +29,88 @@ export async function mockFetch({
...
+    for (const matchingRequest of this.requestList) {
+      if (
+        matchingRequest.url === request.url &&
+        matchingRequest.method === request.method
+      ) {
+        if (request.body === undefined || request.body === null) {
+          return matchingRequest;
+        } else if (matchingRequest.body === request.body) {
+          return matchingRequest;
+        }
```
paulomarg, line 102: This would be a test setup error, but what if we try to find a request that should have a body but doesn't? Could we check that the body matches regardless of it being null?

```suggestion
        if (!request.body && !matchingRequest.body) {
          return matchingRequest;
        } else if (matchingRequest.body === request.body) {
          return matchingRequest;
        }
```

src/clients/http_client/__tests__/http_client.test.ts
mkevinosullivan: Now using `canonicalizeHeaders()` when queuing a mocked response.

src/clients/http_client/__tests__/http_client.test.ts
mkevinosullivan: Now using `canonicalizeHeaders()` when queuing a mocked response.

src/clients/http_client/__tests__/http_client.test.ts
mkevinosullivan: Now using `canonicalizeHeaders()` when queuing a mocked response.

src/clients/http_client/__tests__/http_client.test.ts
mkevinosullivan: Resolved while pairing with @paulomarg

PR #485, opened 2022-09-09
[WIP] Refactor Webhooks to new isomorphic structure

mkevinosullivan:

Refactoring webhooks sub-component to connect into `shopifyApi` object.


src/webhooks/index.ts
```diff
@@ -1,8 +1,10 @@
-import {WebhooksRegistry} from './registry';
+import {ConfigInterface} from '../base-types';
 
-const ShopifyWebhooks = {
-  Registry: WebhooksRegistry,
-};
+import {gdprTopics, webhooksRegistry} from './registry';
```
paulomarg, line 3: I wonder if we should export the GDPR topics as an enum out of the main export, with the other types?

src/webhooks/__tests__/registry.test.ts
```diff
@@ -54,10 +54,10 @@ async function genericWebhookHandler(
 
 const TEST_WEBHOOK_ID = 'gid://shopify/WebhookSubscription/12345';
 
-describe('ShopifyWebhooks.Registry.register', () => {
+describe('shopify.webhooks.registry.register', () => {
   beforeEach(async () => {
-    config.apiVersion = ApiVersion.Unstable;
-    ShopifyWebhooks.Registry.webhookRegistry = {};
+    shopify.config.apiVersion = ApiVersion.Unstable;
+    shopify.webhooks.registry.webhookRegistry = {};
```
paulomarg, line 60: If we're creating a new instance for each test, do we still need to do this? Just curious 🙂

src/webhooks/__tests__/registry.test.ts
```diff
@@ -195,58 +195,58 @@ describe('ShopifyWebhooks.Registry.register', () => {
...
 
     // Update the second handler and make sure we still have the two of them
-    fetchMock.mockResponseOnce(JSON.stringify(webhookCheckResponse));
-    fetchMock.mockResponseOnce(JSON.stringify(successUpdateResponse));
+    queueMockResponse(JSON.stringify(webhookCheckResponse));
+    queueMockResponse(JSON.stringify(successUpdateResponse));
     webhook.path = '/webhooks/new';
-    await ShopifyWebhooks.Registry.register(webhook);
-    expect(Object.keys(ShopifyWebhooks.Registry.webhookRegistry)).toHaveLength(
+    await shopify.webhooks.registry.register(webhook);
```
paulomarg, line 243: Personally, I feel like the interface would be a bit cleaner if we exported the registry _as_ `webhooks`, so we can call e.g. `shopify.webhooks.register()`.

src/webhooks/registry.ts
```diff
@@ -199,270 +199,271 @@ function buildQuery(
...
+  const WebhooksRegistry: RegistryInterface = {
+    webhookRegistry: {},
 
-  addHandlers(handlers: AddHandlersProps): void {
-    for (const topic in handlers) {
-      if ({}.hasOwnProperty.call(handlers, topic)) {
-        WebhooksRegistry.addHandler(topic, handlers[topic]);
+    addHandler(
+      topic: string,
+      {path, webhookHandler}: WebhookRegistryEntry,
```
paulomarg, line 217: We should consider refactoring these to take in objects too for consistency.

paulomarg (commented):
LGTM on a quick skim.

I wonder if we want to break up the registry object into separate functions with their own `create` calls like we do elsewhere, but they have the shared registry object, so I'm not sure it behaves exactly like the other calls. Just thought it'd be worth discussing.

src/webhooks/__tests__/registry.test.ts
```diff
@@ -293,93 +294,153 @@ describe('ShopifyWebhooks.Registry.register', () => {
...
+    shopify.webhooks.addHandler({
+      topic: 'PRODUCTS',
+      registryEntry: {
+        path: '/webhooks',
+        webhookHandler: genericWebhookHandler,
+      },
     });
     const app = express();
-    app.post('/webhooks', ShopifyWebhooks.Registry.process);
+    app.use(parseRawBody);
```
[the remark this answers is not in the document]
mkevinosullivan: I've refactored this to use a single instance of the express app.

src/webhooks/__tests__/registry.test.ts
```diff
@@ -293,93 +294,153 @@ describe('ShopifyWebhooks.Registry.register', () => {
...
 
     const app = express();
-    app.post('/webhooks', ShopifyWebhooks.Registry.process);
+    app.use(parseRawBody);
+    app.post('/webhooks', async (req, res) => {
+      let errorThrown = false;
+      let response: NormalizedResponse = {
+        statusCode: 200,
+        statusText: 'OK',
+      };
```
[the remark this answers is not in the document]
mkevinosullivan: Removed due to other refactoring.

src/webhooks/__tests__/registry.test.ts
```diff
@@ -365,277 +376,129 @@ describe('shopify.webhooks.process', () => {
...
-        await shopify.webhooks.process({
-          rawBody: (req as any).rawBody,
-          rawRequest: req,
-          rawResponse: res,
-        });
-      } catch (error) {
-        errorThrown = true;
-        res.writeHead(error.response.code);
-        res.end();
-        expect(error).toBeInstanceOf(ShopifyErrors.InvalidWebhookError);
```
paulomarg, line 403: Should we also include the exception in the response data, so that we can keep running this check?

src/webhooks/__tests__/registry.test.ts
```diff
@@ -365,277 +376,129 @@ describe('shopify.webhooks.process', () => {
...
-        errorThrown = true;
-        res.writeHead(error.response.code);
-        res.end();
-      }
-      expect(errorThrown).toBeFalsy();
-      expect(response.statusCode).toBe(200);
-    });
-
-    await request(app)
+    const response = await request(app)
```
paulomarg, line 382: Loving how this cleaned up this test, didn't realize we had so many instances of the test express app.

src/webhooks/registry.ts
```diff
@@ -354,11 +356,71 @@ export interface WebhookProcessParams extends AdapterArgs {
   rawBody: string;
 }
 
+const statusTextLookup: {[key: string]: string} = {
```
paulomarg, line 359: While I'm a fan of this, I wonder if it'd be worth making it more generic, since it seems pretty useful all around. Not sure if we'd use it anywhere else though!

src/webhooks/registry.ts
```diff
@@ -354,11 +356,71 @@ export interface WebhookProcessParams extends AdapterArgs {
...
+
+  return retVal;
+}
+
 export function createProcess(config: ConfigInterface) {
   return async function process({
     rawBody,
     ...adapterArgs
-  }: WebhookProcessParams): Promise<NormalizedResponse> {
+  }: WebhookProcessParams): Promise<void> {
```
paulomarg, line 423: Should we not be returning `AdapterResponse` here, as the outcome of `abstractConvertResponse`? Will this work on workers?

src/webhooks/registry.ts
```diff
@@ -367,105 +429,52 @@ export function createProcess(config: ConfigInterface) {
...
-      },
-    );
+    response.statusText = statusTextLookup[response.statusCode];
+    abstractConvertResponse(response, adapterArgs);
+    if (!isOK(response)) {
+      throw new ShopifyErrors.InvalidWebhookError({
+        message: errorMessage,
+        response,
+      });
+    }
```
paulomarg, line 475: As per above, would this be. Also, I think we may need an `await` when converting?

```suggestion
    response.statusText = statusTextLookup[response.statusCode];
    const returnResponse = await abstractConvertResponse(response, adapterArgs);
    if (!isOK(response)) {
      throw new ShopifyErrors.InvalidWebhookError({
        message: errorMessage,
        response,
      });
    }
    
    return returnResponse;
```

src/error.ts
```diff
@@ -84,32 +86,16 @@ export class SessionNotFound extends ShopifyError {}
...
 
-interface InvalidWebhookParams extends HttpResponseData {
+interface InvalidWebhookParams {
   message: string;
-  cause?: Error;
+  response: AdapterResponse;
 }
 export class InvalidWebhookError extends ShopifyError {
-  readonly response: HttpResponseData;
+  readonly response: AdapterResponse;
```
paulomarg, line 94: I wonder if this typing would work for node - in that case `AdapterResponse` is `undefined`, so we wouldn't be surfacing any of the error information. I think we may want to keep this type the way it was before?

paulomarg (commented):
Nice, I think we're really close here!

src/webhooks/registry.ts
```diff
@@ -199,45 +206,49 @@ function buildQuery(
...
+export function getTopics(): string[] {
+  return Object.keys(webhookRegistry);
+}
 
-  getTopics(): string[] {
-    return Object.keys(WebhooksRegistry.webhookRegistry);
-  },
+export function createRegister(config: ConfigInterface) {
+  const HttpClient = createHttpClientClass(config);
+  const GraphqlClient = createGraphqlClientClass({config, HttpClient});
```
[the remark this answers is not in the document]
paulomarg: Fair point!

src/base-types.ts
```diff
@@ -61,7 +63,14 @@ export enum ShopifyHeader {
   Topic = 'X-Shopify-Topic',
   Domain = 'X-Shopify-Shop-Domain',
 }
+
 export enum ClientType {
   Rest = 'rest',
   Graphql = 'graphql',
 }
+
+export const gdprTopics: string[] = [
```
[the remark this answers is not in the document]
paulomarg: Yeah I guess that's fair, let's just keep it as is.

src/webhooks/registry.ts
mkevinosullivan: 🤦🏻‍♂️ Right, we have to return it for the Workers scenario

src/error.ts
paulomarg: Ah no that's fair - we want to provide the actual response in the object, but for node it's awkward because there'll be nothing there. Would it make sense to _also_ return a bit of information on the request to make it easier to debug errors?

src/webhooks/registry.ts
mkevinosullivan: The only other place in non-test code that we could use something like it is https://github.com/Shopify/shopify-api-node/blob/3e251da3b9d60810a9b33cef7bedafd58f10f8cd/src/auth/oauth/oauth.ts#L80-L81

src/webhooks/__tests__/registry.test.ts
mkevinosullivan: The check is still being run, just inside the post `/webhooks` handler - https://github.com/Shopify/shopify-api-node/blob/aabbbcc8ef0f606499131de52ca660796c19e6d2/src/webhooks/__tests__/registry.test.ts#L350-L367

(see Ln 363)

src/error.ts
mkevinosullivan: So, also include the original `AdapterRequest`?

paulomarg: Something like that, though I'm not sure if there'll ever be any data there that's worth returning for webhooks. Maybe we should just keep what you did here and fix forward if we find any information we want to return going forward.

paulomarg (approved):

PR #488, opened 2022-09-12
Make OAuth isomorphic

paulomarg:

Now that all of the pieces are in place, we need to make sure the app can perform OAuth on multiple runtimes.

Fixing any issues that arose from running OAuth in a worker app.

### DIFF
--- a/src/auth/oauth/oauth.ts
+++ b/src/auth/oauth/oauth.ts
@@ -20,9 +20,9 @@ import {
   abstractConvertResponse,
   abstractConvertHeaders,
   AdapterResponse,
+  AdapterHeaders,
   Cookies,
   NormalizedResponse,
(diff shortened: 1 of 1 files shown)


src/__tests__/test-helper.ts
```diff
@@ -110,7 +110,7 @@ export async function setSignedSessionCookie({
   });
 
   // eslint-disable-next-line require-atomic-updates
-  request.headers.Cookie = cookies.toHeaders().join(',');
+  request.headers.Cookie = cookies.toHeaders().join(';');
```
paulomarg, line 113: As per https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cookie#directives

mkevinosullivan (approved):

PR #496, opened 2022-09-16
Refactor rest resources

paulomarg:

Now that we have isomorphic client support and the non-global config all set up, we can add the REST resources back.

Updating the base REST client to work with the new format, and adding the means to import the resources into the library.

Link to review excluding the resources themselves:

`https://github.com/Shopify/shopify-api-node/pull/496/commits/cf9b98d3051f80e8f6920f4275a1df2a596fd071`

Complete example including session storage, REST resources and the CF worker import:
```ts
import "@shopify/shopify-api/adapters/cf-worker";
import { shopifyApi } from "@shopify/shopify-api";
import { MemorySessionStorage } from "@shopify/shopify-api/session-storage/memory";
import { restResources } from "@shopify/shopify-api/rest/admin/2022-07";

const shopify = shopifyApi({
  ...
  sessionStorage: new MemorySessionStorage(),
  restResources,
});

const products = await shopify.rest.Product.all({session, limit: 10});
```

This provides full typing / intellisense support in VSCode:

image


mkevinosullivan (approved):
(based on the initial commit, not the generated files)

PR #504, opened 2022-09-21
Make the billing component isomorphic

paulomarg:
This PR is re-introducing the billing component to the isomorphic library, and ensuring it works on CF workers.


src/billing/request_payment.ts
```diff
@@ -1,78 +1,102 @@
...
+      });
+    }
 
-  const client = new GraphqlClient(session.shop, session.accessToken);
+    return data.confirmationUrl;
+  };
+}
 
+async function requestRecurringPayment({
+  config,
```
mkevinosullivan, line 82: Again, this `config` refers to `BillingConfig`, not a `ConfigInterface` config, which is kinda confusing given that both are referenced in the one file.

src/billing/request_payment.ts
```diff
@@ -94,28 +118,20 @@ async function requestRecurringPayment(
...
-    throw new BillingError({
-      message: 'Attempted to request single payment without billing configs',
-      errorData: [],
-    });
-  }
-
-  const client = new GraphqlClient(session.shop, session.accessToken);
-
+async function requestSinglePayment({
+  config,
```
mkevinosullivan, line 122: `BillingConfig` vs `ConfigInterface`

mkevinosullivan (approved):

PR #506, opened 2022-09-21
Adds "Stale" and "Waiting for Response" labels

klenotiw:

Working on https://github.com/Shopify/first-party-library-planning/issues/458

This PR adds two new labels "Stale" and "Waiting for Response". Stale will be added to any issue that hasn't had activity for 90 days. If an issue is labeled as "Stale" and dose not have activity for an additional 14 days it will be closed automatically. The "Waiting for Response" label is added manually to an issue, if one of our devs is waiting for clarifications on a ticket. If the ticket doesn't see activity for 7 days it will be closed automatically. If it does see activity then the label is removed automatically.

## Type of change

## Checklist

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,31 @@
+name: Close inactive issues
+on:
+  schedule:
+    - cron: "30 1 * * *"
+
+jobs:
+  close-issues:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: actions/stale@v5
+        with:
+          days-before-issue-stale: 90
+          days-before-issue-close: 14
+          stale-issue-label: "Stale"
+          stale-issue-message: >
+            This issue is stale because it has been open for 90 days with no activity. It will be closed if no further action occurs in 14 days. 
+          close-issue-message: |
+            We are closing this issue because it has been inactive for a few months. 
+            This probably means that it is not reproducible or it has been fixed in a newer version. 
+            If it’s an enhancement and hasn’t been taken on since it was submitted, then it seems other issues have taken priority.
+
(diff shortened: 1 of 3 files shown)


klenotiw:
I can see there is a `stale.yml` file in this repo but I am not sure how exactly it works. We can remove my `stale.yml` workflow if the other one works fine.

mkevinosullivan:
> I can see there is a `stale.yml` file in this repo but I am not sure how exactly it works. We can remove my `stale.yml` workflow if the other one works fine.

I committed it originally but it didn't work ... never went back to make it work 🤦🏻‍♂️    It can be safely deleted in preference for yours!

.github/workflows/stale.yml
```diff
@@ -0,0 +1,31 @@
...
+jobs:
+  close-issues:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: actions/stale@v5
+        with:
+          days-before-issue-stale: 90
```
mkevinosullivan, line 15: So this deletes any issues that haven't had any activity for the last 90 days.  During one of our meetings, I though we'd decided to make Jan 1, 2022 the cutoff date.  Can this work with dates or only days?

klenotiw: Good point! We changed it for the `shopify_app` repo to 90 days but I can change it to the start of the year for the rest of the repos. I couldn't find any obvious way to do it by date instead of number of days, so I think the best thing to do would change the number of days to like 264 which would be the start of the year.

mkevinosullivan: Yeah, was trying to figure out if we could set it with `days-before-issue-stale: ${{ some expression code here }}` but it's probably more hassle than the effort required!

What I was going to do was set it to `-1` and then have a script that sets the `Stale` label on issues that haven't had activity since date X

Issue #448, opened 2022-08-05
Improve `sanitizeShop` to work with Spin


hannachen:
## Overview/summary

We're expecting `shop` domain to be limited to `myshopify.com` and `myshopify.io`, but shop urls looks different on spin: `spin.dev`. This means we are unable to develop and test apps against a spin environment.

https://github.com/Shopify/shopify-api-node/blob/54b8cfe51682254f708ec623a97a9489459b20ad/src/utils/shop-validator.ts#L15

## Motivation

Some apps need to be tested or developed end-to-end with changes in spin environment.

### Area

- [ ] Add any relevant `Area: <area>` labels to this issue

---

## Checklist

- [x] I have described this enhancement in a way that is actionable (if possible)

PR #514, opened 2022-09-27
Support spin stores by default

Arkham:

We want to automatically support stores that are hosted on spin environments, as it makes easier to test tools internally.

## Type of change

## Checklist

### DIFF
--- a/src/utils/__tests__/shop-validator.test.ts
+++ b/src/utils/__tests__/shop-validator.test.ts
@@ -6,6 +6,7 @@ const VALID_SHOP_URL_1 = 'someshop.myshopify.com';
 const VALID_SHOP_URL_2 = 'devshop.myshopify.io';
 const VALID_SHOP_URL_3 = 'test-shop.myshopify.com';
 const VALID_SHOP_URL_4 = 'dev-shop-.myshopify.io';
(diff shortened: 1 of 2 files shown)


mkevinosullivan:
Proposed fix for #448

mkevinosullivan:
Hi @Arkham, instead of changing this code, you can add a custom shop domain to the `Context` which will automatically be checked in the `sanitizeShop` method (https://github.com/Shopify/shopify-api-node/blob/main/src/utils/shop-validator.ts#L16-L23).

```ts
Shopify.Context.initialize({
  API_KEY,
  API_SECRET_KEY,
  SCOPES: [SCOPES],
  HOST_NAME: HOST.replace(/https?:\/\//, ""),
  IS_EMBEDDED_APP: true,
  API_VERSION: ApiVersion.July22,
  CUSTOM_SHOP_DOMAINS: '.*\\.spin\\.dev',
});
```

mkevinosullivan:
We intentionally wanted to avoid any reference to `spin.dev` within the library itself, although I think it's kinda public at this stage that we use a cloud dev environment.

The only reference to this capability is buried, as the last line, in this [doc](https://github.com/Shopify/shopify-api-node/blob/main/docs/usage/utils.md), which itself isn't linked from anywhere, or none that I could find.  That's an improvement I'll add to our project board.

PR #520, opened 2022-09-29
Add E2E HTTP client tests

paulomarg:

Since we can now support multiple runtimes, we want to increase our confidence that we're not introducing errors by only testing using a mock adapter.

To increase our coverage, this PR adds a suite of generic tests to exercise the `HttpClient` class, and sets up actual internal servers on node and CF workers.

By running these actual servers, we can test that the actual implementations work properly, as far as making requests goes.


src/auth/oauth/oauth.ts
```diff
@@ -24,6 +24,8 @@ import {
 } from '../../runtime/http';
 
 import {
+  SESSION_COOKIE_NAME,
```
paulomarg, line 27: I noticed a complaint about a circular dependency here, so I moved these values to the types file.

src/clients/http_client/__e2etests__/create_shopify_object.ts
```diff
@@ -0,0 +1,15 @@
+import {MemorySessionStorage} from '../../../session-storage/memory';
```
mkevinosullivan, line 1: Since this file is no longer creating the shopify object, merely the config, should the file be renamed?

PR #521, opened 2022-09-30
Use md diffs in migration guide

paulomarg:
This just replaces the `before`/`after` code blocks in the migration guide with diff blocks, which hopefully reduces the mental load of finding the actual changes to be made.

The question is: does it?


paulomarg:
> 

I had the same feeling but didn't want to bias anyone. I think we should abandon this and stick with the current format, which is A LOT easier to copy-paste too.

PR #541, opened 2022-10-06
Fix : remove surrounding curly brackets because

bmatthieuedn:

Fixes #0000 

## Type of change

## Checklist

### DIFF
--- a/src/auth/session/storage/README.md
+++ b/src/auth/session/storage/README.md
@@ -28,12 +28,12 @@ Shopify.Context.initialize({
 // OR
 
 Shopify.Context.initialize({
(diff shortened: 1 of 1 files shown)


jw-soh:
Hi can someone please review this?

cquemin (commented):
We will soon release v6 of `shopify/shopify-api` where session management has changed quite a bit, in particular where named parameter are now mandatory for methods calls and session storage is now externalized outside the lib. Also the PR doesn't really say why the curly braces have been removed, neither does the commit messages. This could explain why it has taken so long to be reviewed.
So it is difficult to know if there will be issues with v6 because of curly braces. Keen to discuss @jw-soh @bmatthieuedn

PR #558, opened 2022-10-09
Improve RedisSessionStorage.findSessionsByShop:  Do Not Fail if Non-Session Data Exists in Redis 

devstein:

Right now, `findSessionsByShop` throws an exception if the Redis instance has non-session data that isn't parse-able by `JSON.parse` or `sessionFromEntries`

This PR skips values (`rawResult`) stored in Redis that are not related to shop sessions.

## Type of change

## Checklist

### DIFF
--- a/src/auth/session/storage/redis.ts
+++ b/src/auth/session/storage/redis.ts
@@ -92,12 +92,17 @@ export class RedisSessionStorage implements SessionStorage {
       if (!rawResult) continue;
 
       try {
-
-      const session = sessionFromEntries(JSON.parse(rawResult));
-      if (session?.shop === cleanShop) results.push(session);
-      } 
-      // do nothing if the rawResult is not a parse-able session
-      catch {}
+        const session = sessionFromEntries(JSON.parse(rawResult));
+        if (session?.shop === cleanShop) results.push(session);
+      } catch (err: unknown) {
+        // do nothing if the rawResult is not a parse-able session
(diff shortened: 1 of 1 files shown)


devstein:
Hi 👋 , any recommendations for testing? This edge-case is implementation specific and doesn't fit into the `SessionStorage` interface (i.e as a new test in battery-of-tests)

devstein:
Signed the CLA

src/auth/session/storage/redis.ts
```diff
@@ -85,13 +85,19 @@ export class RedisSessionStorage implements SessionStorage {
...
       if (!rawResult) continue;
 
+      try {
+
       const session = sessionFromEntries(JSON.parse(rawResult));
-      if (session.shop === cleanShop) results.push(session);
+      if (session?.shop === cleanShop) results.push(session);
+      } 
+      // do nothing if the rawResult is not a parse-able session
+      catch {}
```
cquemin, line 100: Should we catch the Json Error raised here or any more relevant parsing exception? I am conscious of hiding other exceptions that we may want to let through if they occur. I know this would require some custom code in order to get the type of the exception or re throw it if it is not what we want.
Not sure if this is very JS idiomatic but `catch all` blocks always scare me :)

devstein: Narrowing error types isn't elegant in JS, but gave it go. I agree that the `catch all` is scary

devstein (commented):
@cquemin Updated. Let me know what you think

PR #560, opened 2022-10-10
Add `overwrite` and `path` scope STATE_COOKIE_NAME cookie

deldrid1:

My OAuth workflow roughly follows https://rsirokov.medium.com/shopify-app-online-and-offline-access-modes-8a8c5ecd928b.  I am attempting to authenticate offline first at Custom app install time, then get online credentials for the user without needing to go through two user facing authorization screens.  Basically, you hit the following routes on my side, with the necessary Shopify OAuth endpoints in the middle

```
  // /shopify/auth/toplevel
  // /shopify/auth/offline - initiate OAuth complete with user facing acceptance for offline tokens
  // /shopify/auth/offline/callback - get and store offline tokens
  // /shopify/auth/online - redirect to initiate OAuth for online token (no user screen shown as long as scopes match)
  // /shopify/auth/online/callback - get online token
```

In migrating up to WIP v6, I was running into an issue where my OAuth query parameter state was not matching the state stored in the `shopify_app_state` cookie sent to my server, despite it looking correct in the Inspector.  Inspecting network requests didn't really reveal why this mismatch was happening - all the headers and order of operations looked sane to me.

To make things more fun, once I got a 400 - `Invalid OAuth callback` sent to the browser, a refresh would get things working 🤦‍♂️

I'm not sure what change happened from 5 -> 6, or why Chrome and Safari are both being confused with the cookie changing going on, but the fix that managed to kick things loose was to ensure that the `shopify_app_state` cookie is both overwritten (I am hosting this in a serverless environment so no guarantee that the `nonce` won't change between OAuth kickoffs - this makes sense to do for everyone I think?) AND is scoped to the specific path it is being used for

## Type of change

## Checklist

### DIFF
--- a/src/auth/oauth/oauth.ts
+++ b/src/auth/oauth/oauth.ts
@@ -63,6 +63,8 @@ export function createBegin(config: ConfigInterface) {
       expires: new Date(Date.now() + 60000),
       sameSite: 'lax',
       secure: true,
+      overwrite: true,
+      path: callbackPath
     });
 
     const query = {


deldrid1:
CLA signed if someone with access can rerun the workflow.

paulomarg (approved):
Tried it out and it makes sense to me, it enables apps with different paths to more easily isolate the state cookie.

The big difference here is that we started saving the state to the cookie, rather than a session id.

Issue #465, opened 2022-08-23
Create a separate package for Session Storage Adapters


Praveen2106:
## Overview
Move the existing Session Storage Adapters to a separate package. Managing Sessions is not the Core Functionality of this Library. 

It makes sense to provide Memory Session Storage for testing purposes and an extendable API for other Databases to use in production.

## Type

- [x] Changes to existing features

## Motivation

> What inspired this feature request? What problems were you facing?
We faced issues with the build due to compatibility with the packages used for Session Storage. Reducing the number of unused dependencies resolves many Build Issues reported here.

murphpdx:
We would also like this feature. We have our own solution for this and do not want to have an additional parallel system.

maktouch:
Our build is failing because of a node-gyp error, and it's because shopify-api is installing sqlite, which we don't need. At the very least, the database drivers should be devDeps or peerDeps... to make them actually dependencies is not efficient at all, and causes a lot of trouble. 

```
"dependencies": {
    "@shopify/network": "^1.5.1",
    "@types/jsonwebtoken": "^8.5.0",
    "@types/node-fetch": "^2.5.7",
    "@types/supertest": "^2.0.10",
    "cookies": "^0.8.0",
    "jsonwebtoken": "^8.5.1",
    "mongodb": "^4.5.0", // ===========> to remove
    "mysql2": "^2.3.3", // ===========> to remove
    "node-fetch": "^2.6.1",
    "pg": "^8.7.3", // ===========> to remove
    "redis": "^4.1.0", // ===========> to remove
    "sqlite3": "^5.0.8", // ===========> to remove
    "tslib": "^2.0.3",
    "uuid": "^8.3.1"
  },
```

dxptqhtlutehvlyxcmtg:
+1 for us too. These dependencies should be opt-in, not mandatory. Some of them require special configuration to build properly in certain situations. Users shouldn't have to be forced to do that for something they're not even using.

dxptqhtlutehvlyxcmtg:
For anyone like us bundling their server code with Webpack (NestJs actually) and don't mind keeping `node_modules` around, we are able to build v5 of this package with the help of `webpack-node-externals` on Node 16:

```javascript
// webpack.config.js
const nodeExternals = require("webpack-node-externals");

const webpackConfig = {
  externals: [
    nodeExternals(),
    {
      "@shopify/shopify-api": "commonjs2 @shopify/shopify-api",
    },
  ],
};

[...]

Issue #410, opened 2022-06-23
Unable to use @shopify/shopify-api on lambda aws function


renanbronchart:
# Issue summary

Write a short description of the issue here ↓

I tried to use @shopify/shopify-api on lambda aws.
But when my lambda function I have an error with this package and `napi-v6-linux-glibc-x64/node_sqlite3.node`.

On local in my macOs, with nodev14, the function works. But when it is executed on lambda, it doesn't work.

## Expected behavior

@shopify/shopify-api should work and I don't want to have this error.

## Actual behavior

I receive this error on lambda : 
```
{
    "errorType": "Runtime.ImportModuleError",
    "errorMessage": "Error: Cannot find module '/var/task/node_modules/sqlite3/lib/binding/napi-v6-linux-glibc-x64/node_sqlite3.node'\nRequire stack:\n- /var/task/node_modules/sqlite3/lib/sqlite3-binding.js\n- /var/task/node_modules/sqlite3/lib/sqlite3.js\n- /var/task/node_modules/@shopify/shopify-api/dist/auth/session/storage/sqlite.js\n- /var/task/node_modules/@shopify/shopify-api/dist/context.js\n- /var/task/node_modules/@shopify/shopify-api/dist/index.js\n- /var/task/utils.js\n- /var/task/index.js\n- /var/runtime/UserFunction.js\n- /var/runtime/index.js",
    "stack": [
        "Runtime.ImportModuleError: Error: Cannot find module '/var/task/node_modules/sqlite3/lib/binding/napi-v6-linux-glibc-x64/node_sqlite3.node'",
        "Require stack:",
        "- /var/task/node_modules/sqlite3/lib/sqlite3-binding.js",
        "- /var/task/node_modules/sqlite3/lib/sqlite3.js",
        "- /var/task/node_modules/@shopify/shopify-api/dist/auth/session/storage/sqlite.js",
        "- /var/task/node_modules/@shopify/shopify-api/dist/context.js",
        "- /var/task/node_modules/@shopify/shopify-api/dist/index.js",
        "- /var/task/utils.js",
        "- /var/task/index.js",
        "- /var/runtime/UserFunction.js",
        "- /var/runtime/index.js",
        "    at _loadUserApp (/var/runtime/UserFunction.js:202:13)",
        "    at Object.module.exports.load (/var/runtime/UserFunction.js:242:17)",
        "    at Object.<anonymous> (/var/runtime/index.js:43:30)",
[...]

Issue #400, opened 2022-06-02
Not able to install shopify-node-api


gaurav-pachar:
# Issue summary

unable to install shopify-api with,  npm i @shopify/shopify-api 

## Expected behavior

dependency should get installed

## Actual behavior

while executing npm i @shopify/shopify-api, the process gets interrupted at
 ```reify:@shopify/shopify-api: timing reifyNode:node_modules/@redis/client Completed in 361ms```

the whole error log after interruption is:
```
npm i @shopify/shopify-api
npm ERR! code 1
npm ERR! path /Users/gauravpachar/untitled folder/node_modules/sqlite3
npm ERR! command failed
npm ERR! command sh -c node-pre-gyp install --fallback-to-build
npm ERR! CC(target) Release/obj.target/nothing/../node-addon-api/nothing.o
npm ERR!   LIBTOOL-STATIC Release/nothing.a
npm ERR!   ACTION deps_sqlite3_gyp_action_before_build_target_unpack_sqlite_dep Release/obj/gen/sqlite-autoconf-3380400/sqlite3.c
npm ERR!   TOUCH Release/obj.target/deps/action_before_build.stamp
npm ERR!   CC(target) Release/obj.target/sqlite3/gen/sqlite-autoconf-3380400/sqlite3.o
npm ERR!   LIBTOOL-STATIC Release/sqlite3.a
npm ERR!   CXX(target) Release/obj.target/node_sqlite3/src/backup.o
npm ERR! Failed to execute '/opt/homebrew/Cellar/node/18.2.0/bin/node /opt/homebrew/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js build --fallback-to-build --module=/Users/gauravpachar/untitled folder/node_modules/sqlite3/lib/binding/napi-v6-darwin-unknown-arm64/node_sqlite3.node --module_name=node_sqlite3 --module_path=/Users/gauravpachar/untitled folder/node_modules/sqlite3/lib/binding/napi-v6-darwin-unknown-arm64 --napi_version=8 --node_abi_napi=napi --napi_build_version=6 --node_napi_label=napi-v6' (1)
npm ERR! node-pre-gyp info it worked if it ends with ok
npm ERR! node-pre-gyp info using node-pre-gyp@1.0.9
npm ERR! node-pre-gyp info using node@18.2.0 | darwin | arm64
[...]

PR #561, opened 2022-10-10
Remove node default storage

deldrid1:

Address some pain points associated with #465, #410, and #400.  Since v6 is a MAJOR update, it makes sense to make this change now, require a tiny bit more of extra configuration, and make currently broken deployment scenarios possible (like AWS Lambda!), without any crazy hacks.

> Note: there is probably some additional documentation like the main README to update for this - I only addressed the Migration Guide.

Instead of cramming `SQLiteSessionStorage` down people's throats as the default Session Storage (regardless of whether it was being used), consumers of the library now have to bring their own (by importing one in the library that exists or writing their own). 

## Type of change

## Checklist


paulomarg:
I think this is a valid point! I'll bring this to the team and we'll see how we want to tackle this issue.

PR #563, opened 2022-10-11
Default to console calls for logFunction

paulomarg:

Internally, whenever we called `logFunction`, we had to check if it was set first, which needlessly bloated the code.

Default the `logFunction` parameter to a noop function that just resolves a promise, so that we don't have to check before calling it.

### DIFF
--- a/adapters/__e2etests__/utils.ts
+++ b/adapters/__e2etests__/utils.ts
@@ -33,4 +33,5 @@ export const config: ConfigInterface = {
   isEmbeddedApp: true,
   isPrivateApp: false,
   sessionStorage: new MemorySessionStorage(),
+  logFunction: () => Promise.resolve(),
 };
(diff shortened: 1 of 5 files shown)


src/clients/http_client/http_client.ts
```diff
@@ -293,15 +293,13 @@ export function createHttpClientClass(
...
-          } else {
-            console.warn('API Deprecation Notice:', deprecation);
-          }
+          console.warn('API Deprecation Notice:', deprecation);
+
+          const stack = new Error().stack;
+          const log = `API Deprecation Notice ${new Date().toLocaleString()} : ${JSON.stringify(
+            deprecation,
+          )}\n    Stack Trace: ${stack}\n`;
+          await config.logFunction(LogSeverity.Warning, log);
```
mkevinosullivan, line 302: The previous code did `console.warn` if there was no log function, otherwise the log function was called.  This new code is doing both.  IIRC the original implementation was to reduce the number of warnings (noise) to the console and have them sent to a log somewhere, no?

mkevinosullivan: Maybe the default implementation should just be a call to `console.warn`?

paulomarg: That's a great callout - maybe we should have a function that just logs for info, warns for warnings, etc... Much cleaner!

PR #566, opened 2022-10-13
Fix Webpack 5 bundling error

deldrid1:
Webpack 5 was throwing an error when bundling
```
Module not found: Error: Default condition should be last one
```

One-line to change the order of the `exports` in `package.json` so that webpack stops complaining.

Per https://github.com/bcakmakoglu/revue-draggable/issues/223#issue-1124523112:
> There is a rule in webpack 'enhanced-resolve' causing this error [here](https://github.com/webpack/enhanced-resolve/blob/02d99a2e6852adaf43fecfc3fed14d8d5c8df10b/lib/util/entrypoints.js#L455)
>
> It can be easily fixed by moving "default" entry in package.json to be the last entry

## Type of change

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,9 +6,9 @@
   "types": "./lib/index.d.ts",
   "exports": {
     ".": {
-      "default": "./lib/index.js",
       "require": "./lib/index.js",
-      "types": "./lib/index.d.ts"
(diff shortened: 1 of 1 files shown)


paulomarg (approved):
Makes sense to me, good catch!

PR #568, opened 2022-10-18
Update session-utils.ts

ashiqueacodez:
I'm getting the following error when I use Shopify.Session.RedisSessionStorage(REDIS_URL)

TypeError: entries.filter is not a function
@shopify/shopify-api/dist/auth/session/session-utils.js:11:10

Fixes #0000 

## Type of change

## Checklist


cquemin (commented):
We will soon release v6 of `shopify/shopify-api` and session management will be quite different so I wonder if this is worth keeping or not?

mkevinosullivan:
With the release of v6, the structure of the API library has been significantly changed, including the removal of the `SessionStorage` classes to their own separate packages in the `Shopify/shopify-app-js` repo.

Hence, this fix can no longer be merged.

PR #572, opened 2022-10-21
Add optional webhook handler fields

paulomarg:

Currently, there are some fields in the webhook mutations that we don't support in the library.

Adding support for those fields, and enabling the logic for updating existing handlers now that there are fields that might change. We still check that the values themselves haven't changed,

### DIFF
--- a/lib/webhooks/register.ts
+++ b/lib/webhooks/register.ts
@@ -140,7 +140,6 @@ function buildHandlerFromNode(edge: WebhookCheckResponseNode): WebhookHandler {
   switch (endpoint.__typename) {
     case 'WebhookHttpEndpoint':
       handler = {
-        id: edge.node.id,
         deliveryMethod: DeliveryMethod.Http,
         callbackUrl: endpoint.callbackUrl,
         // This is a dummy for now because we don't really care about it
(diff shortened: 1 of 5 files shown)


lib/webhooks/register.ts
```diff
@@ -167,21 +166,25 @@ function buildHandlerFromNode(edge: WebhookCheckResponseNode): WebhookHandler {
...
         pubSubTopic: endpoint.pubSubTopic,
       };
       break;
   }
 
+  // Set common fields
+  handler.id = edge.node.id;
+  handler.includeFields = edge.node.includeFields;
+  handler.metafieldNamespaces = edge.node.metafieldNamespaces;
+  handler.privateMetafieldNamespaces = edge.node.privateMetafieldNamespaces;
```
mkevinosullivan, line 186: Wondering if these could be sorted when stored, so that at least that doesn't need to be done in the `categorizeHandlers` later

mkevinosullivan (approved):

lib/webhooks/register.ts
```diff
@@ -252,6 +266,46 @@ function categorizeHandlers(
...
+    privateMetafieldNamespacesEqual
+  );
+}
+
+function arraysEqual(arr1: any[], arr2: any[]): boolean {
+  if (arr1.length !== arr2.length) {
+    return false;
+  }
+
+  for (let i = 0, len = arr1.length; i < len; i++) {
```
teddyhwang, line 301: nit, I don't think there's any benefit to assign the length to a variable

```suggestion
  for (let i = 0; i < arr1.length; i++) {
```

lib/webhooks/register.ts
```diff
@@ -140,30 +140,40 @@ function buildHandlerFromNode(edge: WebhookCheckResponseNode): WebhookHandler {
...
 
+  // Set common fields
+  handler.id = edge.node.id;
+  handler.includeFields = edge.node.includeFields;
+  handler.metafieldNamespaces = edge.node.metafieldNamespaces;
+
+  // Sort the array fields to make them cheaper to compare later on
+  handler.includeFields?.sort();
+  handler.metafieldNamespaces?.sort();
+  if (handler.deliveryMethod === DeliveryMethod.Http) {
```
teddyhwang, line 173: could we add a comment explaining why `privateMetafieldNamespaces` needs this condition?

lib/webhooks/register.ts
paulomarg: Yeah, for the length of the arrays we're dealing with, re-calculating the length wouldn't really have any performance impact.

PR #574, opened 2022-10-24
Fix real-world webhook issues

paulomarg:

After the webhook refactoring, there were some small issues with the queries we run to manage them.

Rearranges the webhooks so that they work consistently across runtimes.

### DIFF
--- a/adapters/cf-worker/adapter.ts
+++ b/adapters/cf-worker/adapter.ts
@@ -23,10 +23,12 @@ export async function workerConvertRequest(
   for (const [key, value] of request.headers.entries()) {
     addHeader(headers, key, value);
   }
+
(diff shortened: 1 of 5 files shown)


adapters/cf-worker/adapter.ts
```diff
@@ -23,10 +23,12 @@ export async function workerConvertRequest(
   for (const [key, value] of request.headers.entries()) {
     addHeader(headers, key, value);
   }
+
+  const url = new URL(request.url!);
```
paulomarg, line 27: We were inconsistent in how we set up the URL for the normalized object for workers - this removes the hostname and protocol so the "internal" URL is just the path and later.

PR #576, opened 2022-10-25
[Fix] Remove $ to enhance commands copy

riccardogiorato:
Removing $ makes it possible to copy the command more easily.


riccardogiorato:
I have signed CLA but can't rerun the actions

PR #581, opened 2022-11-01
[Fix] Update session storage doc for correct MySQLSessionStorage connection string

jw-soh:

Fixes #0000 

The doc has incorrect syntax in the session storage readme.MD file for MySQL.

  Updated the correct syntax in the doc to save time for developers who would run into the same issue in future if MySQL was their chosen session storage db.


jw-soh:
I have signed CLA but can't rerun the actions

cquemin (commented):
We will soon-ish publish v6 of `shopify/shopify-api` and session management will be quite different. So I am wondering if this changes is worth keeping?

jw-soh:
@cquemin , if the new v6 fixes the issue then all's good! thanks for the response :)

mkevinosullivan:
This change is no longer mergeable, as the structure of the code base has changed significantly with `v6`, including the removal of `SessionStorage` classes to their own packages in the `Shopify/shopify-app-js` repo.

PR #586, opened 2022-11-04
Delete fully removed webhook topics

paulomarg:

With the new webhook registration model, we're now deleting handlers that are no longer being used. However, we were not properly handling topics that were fully removed from the registry - i.e. they exist in Shopify, but not in the app's configs.

Going over leftover topics, and deleting any handlers we have for them.

As a bonus, I'm refactoring `webhooks.register` to take in a session object rather than the session fields we'll need.

### DIFF
--- a/lib/webhooks/__tests__/register.test.ts
+++ b/lib/webhooks/__tests__/register.test.ts
@@ -7,6 +7,7 @@ import {queueMockResponse, shopify} from '../../__tests__/test-helper';
 import {mockTestRequests} from '../../../adapters/mock/mock_test_requests';
 import {queryTemplate} from '../query-template';
 import {TEMPLATE_GET_HANDLERS, TEMPLATE_MUTATION} from '../register';
+import {Session} from '../../session/session';
 import {InvalidDeliveryMethodError} from '../../error';
 
 import * as mockResponses from './responses';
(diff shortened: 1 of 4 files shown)


lib/webhooks/__tests__/register.test.ts
```diff
@@ -291,13 +295,70 @@ describe('shopify.webhooks.register', () => {
...
+      `topic: ${topic}`,
+      {callbackUrl: `"https://test_host_name/webhooks"`},
+    );
+    assertWebhookRegistrationRequest(
+      'eventBridgeWebhookSubscriptionCreate',
+      `topic: ${topic}`,
+      {arn: '"arn:test"'},
+    );
+    assertRegisterResponse({registerReturn, topic, responses});
+  });
```
mkevinosullivan, line 361: I was trying to wrap my head around the return result content from register when there are multiple handlers registered per topic (e.g., HTTP and Event Bridge), so I added in this test.

There's probably more derivatives of this test that should be added, e.g., for a topic with multiple handlers
- what happens when the HTTP handler changes but the EB stays the same
- what happens when one of the handlers for a given topic is deleted
- etc.

mkevinosullivan (changes requested):

lib/webhooks/__tests__/register.test.ts
paulomarg: No objections here!

mkevinosullivan (approved):

PR #597, opened 2022-11-14
Update docs based on feedback

mkevinosullivan:
Update docs based on internal user feedback regarding friction points.


cquemin (approved):
LGTM, lots of work in there!!!

PR #598, opened 2022-11-15
one typo

Amit-Flyy:


mkevinosullivan:
As of v6, the library structure and corresponding documentation has changed substantially, hence this is likely not applicable any more.

PR #599, opened 2022-11-15
Fix auth and utils docs to reflect updated v6 library methods and exports

mkevinosullivan:

- `getEmbeddedAppUrl` used to be in `Shopify.Utils`, now in `shopify.auth`- update docs to reflect the change.
- Add documentation for `nonce` and `safeCompare` in `shopify.auth`
- Add documentation for `validateHmac` and `versionCompatible` in `shopify.utils`

### DIFF
--- a/docs/usage/oauth.md
+++ b/docs/usage/oauth.md
@@ -102,7 +102,7 @@ This method takes an object with the following properties:
 
 | Parameter     | Type              |      Required?       | Default Value | Notes                                                                                          |
 | ------------- | ----------------- | :------------------: | :-----------: | ---------------------------------------------------------------------------------------------- |
-| `isOnline`    | `bool`            |         Yes          |       -       | `true` if the session is online and `false` otherwise. Must match the value from `auth.begin`. |
+| `isOnline`    | `boolean`         |         Yes          |       -       | `true` if the session is online and `false` otherwise. Must match the value from `auth.begin`. |
 | `rawRequest`  | `AdapterRequest`  |         Yes          |       -       | The HTTP Request object used by your runtime.                                                  |
 | `rawResponse` | `AdapterResponse` | _Depends on runtime_ |       -       | The HTTP Response object used by your runtime. Required for Node.js.                           |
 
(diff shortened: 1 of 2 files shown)


docs/usage/oauth.md
```diff
@@ -176,4 +176,57 @@ if (!shopify.config.scopes.equals(session.scope)) {
...
+
+const argResult = shopify.auth.safeCompare(arg1, arg2); // throws SafeCompareError due to argument type mismatch
+```
+
+### Generate a cryptographically random string of digits with `nonce`
+
+`nonce` generates a string of 15 characters that are cryptographically random, suitable for short-lived values in cookies to aid validation of requests/responses.
+
+```ts
+const state = shopify.auth.nonce();
```
cquemin, line 229: I thought this wasnt needed to be exposed anymore?

cquemin (commented):
Looks good on the part you modified.

On the part that you didn't modified:

- [in this part](https://github.com/Shopify/shopify-api-node/blob/2f79ff69902a9ae8d0ed50a905aac0e8fdf1740a/docs/usage/oauth.md#loading-a-session-while-handling-a-request) worth referring to the full session storage implementation and may mention that the session should be stored against their Id as this is the way to retrieve a session from shopify requests.
- where it says `navigate to your ngrok address}/auth` perhaps i would be better referring to the doc on how to start your app locally as they may be using ngrock or cloudflare or maybe even do it locally ?
- [in this part](https://github.com/Shopify/shopify-api-node/blob/2f79ff69902a9ae8d0ed50a905aac0e8fdf1740a/docs/usage/oauth.md#loading-a-session-for-a-background-job) It is not clear how to generate the actual offline session object that will then be stored in my session storage, so I can retrieve it later on by using `getOfflineId`. Is it through `auth.begin` ?

docs/usage/oauth.md
mkevinosullivan: I don't see any reference to it no being needed to be exported ... it currently still is exported ... do you have a reference to that somewhere?

mkevinosullivan:
> [in this part](https://github.com/Shopify/shopify-api-node/blob/2f79ff69902a9ae8d0ed50a905aac0e8fdf1740a/docs/usage/oauth.md#loading-a-session-while-handling-a-request) worth referring to the full session storage implementation and may mention that the session should be stored against their Id as this is the way to retrieve a session from shopify requests.

I'm going to have to defer this until we restructure the repos, i.e., when the session storage stuff is in its own package

> where it says navigate to your ngrok address}/auth perhaps i would be better referring to the doc on how to start your app locally as they may be using ngrock or cloudflare or maybe even do it locally ?

I've updated to wording to be more generic, about using a tunnel url, or localhost, or a cloud platform url, depending on how they run their app.  The doc for how to start their app is not in this API library but in the app template repo.  For the API library repo, we should keep it generic.

> [in this part](https://github.com/Shopify/shopify-api-node/blob/2f79ff69902a9ae8d0ed50a905aac0e8fdf1740a/docs/usage/oauth.md#loading-a-session-for-a-background-job) It is not clear how to generate the actual offline session object that will then be stored in my session storage, so I can retrieve it later on by using getOfflineId. Is it through auth.begin ?

Good catch, I've added a "Note 2" to this section about how to obtain an offline session.

PR #606, opened 2022-11-21
[WIP] [Feature] Pass webhook id to handlers

### LINKED ISSUE #370: [Feature] Access to webhook id in handlers
## Overview

I want to process webhooks in an idempotent way. Shopify provides us with `X-Shopify-Webhook-Id` headers. But these get lost in the webhook process function.

It would be great to have access to the webhook ID in my handlers.

## Type

- [ ] New feature
- [x] Changes to existing features

## Motivation

> What inspired this feature request? What problems were you facing?

I want to process webhooks in an asynchronous way, using background workers. For this, I want to use the webhook ID as the background Job ID.

## Checklist

- [x] I have described this feature request in a way that is actionable (if possible)

jbwyme:

Fixes #370 

Grabs the webhook id from the 'X-Shopify-Webhook-Id' header and passes it to the webhook handlers.

## Type of change

## Checklist

### DIFF
--- a/src/webhooks/registry.ts
+++ b/src/webhooks/registry.ts
@@ -360,6 +360,7 @@ const WebhooksRegistry: RegistryInterface = {
         let hmac: string | string[] | undefined;
         let topic: string | string[] | undefined;
         let domain: string | string[] | undefined;
+        let webhookId: string | string[] | undefined;
         Object.entries(request.headers).map(([header, value]) => {
           switch (header.toLowerCase()) {
             case ShopifyHeader.Hmac.toLowerCase():
@@ -370,6 +371,9 @@ const WebhooksRegistry: RegistryInterface = {
               break;
             case ShopifyHeader.Domain.toLowerCase():
               domain = value;
+              break;
+            case ShopifyHeader.WebhookId.toLowerCase():
+              webhookId = value;
               break;
           }
         });
@@ -383,6 +387,9 @@ const WebhooksRegistry: RegistryInterface = {
         }
         if (!domain) {
           missingHeaders.push(ShopifyHeader.Domain);
+        }
+        if (!webhookId) {
+          missingHeaders.push(ShopifyHeader.WebhookId);
         }
 
         if (missingHeaders.length) {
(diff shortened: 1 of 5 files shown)


jbwyme (commented):
remove unnecessary new line

scottdixon:
Hey @jbwyme, thanks for adding this! 

Checks are failing on `Not all authors have signed a CLA or have a GitHub account associated with their email`.

Can you please sign [Shopify’s CLA](https://cla.shopify.com/). Also are you able to [add a test](https://github.com/Shopify/shopify-api-node/blob/main/src/webhooks/__tests__/registry.test.ts#L339)?

src/webhooks/types.ts
```diff
@@ -8,6 +8,7 @@ type WebhookHandlerFunction = (
   topic: string,
   shop_domain: string,
   body: string,
+  webhookId: string,
```
paulomarg, line 11: I think we should probably make this optional to not break existing apps that are using TS.

src/webhooks/registry.ts
```diff
@@ -384,6 +388,9 @@ const WebhooksRegistry: RegistryInterface = {
         if (!domain) {
           missingHeaders.push(ShopifyHeader.Domain);
         }
+        if (!webhookId) {
+          missingHeaders.push(ShopifyHeader.WebhookId);
+        }
```
paulomarg, line 393: Since this header isn't strictly necessary for validation, I wonder if we want to fail on it missing. That being said, we should always have the header so I suppose there's no harm in checking 🙂.

paulomarg (commented):
This is great, thank you! I think we can merge this once we make the type parameter optional.

jbwyme:
Thanks for the review! I got side tracked but I'll finish out this PR today.

src/webhooks/registry.ts
mkevinosullivan: It's optional in the `WebhookHandlerFunction` type, so it probably shouldn't be mandatory here.  The existing tests aren't passing because of this check

jbwyme:
That's what I get for trying to do the whole PR in the github editor :D

src/webhooks/registry.ts
jbwyme: My 2c: It's optional in the type for backwards compatibility but the expected behavior is the header always exists. Therefore, I'd argue this is _more_ correct. However, it's a minor nit and it probably doesn't matter much either way. Happy to go either direction!

paulomarg (approved):

PR #607, opened 2022-11-22
fix: check if shopify_sessions exists based on db name

### LINKED ISSUE #585: Multiple databases on same MySQL instance with the same session table name causes issues with Shopify.Session.MySQLSessionStorage
# Issue summary

When you have two different databases supporting two different Shopify apps on the same MySQL host instance, the hasSessionTable() function within "/node_modules/@shopify/shopify-api/dist/auth/session/storage/mysql.js" returns incorrectly if you use the same session table name. This is because the query to check if the table exists is selecting the table name from the INFORMATION_SCHEMA table and not taking into account the database name.

## Expected behavior

I should be allowed to have the same session table name in two different databases on the same MySQL instance without any issues while using Shopify.Session.MySQLSessionStorage in Shopify.Context.initialize.

## Actual behavior

If I use the same session table name in a new app connected to the same MySQL database instance but using a different database within that instance, it will assume the session table already exists and try to access it to store/read sessions, but that will fail since the table does not actually exist in the database being used.

1. Launch a single mysql server instance and create two databases within named "sample-app-a" and "sample-app-b".
2. Configure a new shopify-cli app to use Shopify.Session.MySQLSessionStorage with session_table name "shop_session" on database "sample-app-a".
3. Run sample-app-a in dev mode using `npm run dev`. (This should create the "shop_session" table in "sample-app-a" database)
5. Configure a second new shopify-cli app to use Shopify.Session.MySQLSessionStorage with session_table name "shop_session" on database "sample-app-b
6. Run sample-app-b in dev mode using `npm run dev`. (This should, but does not, create the "shop_session" table in "sample-app-b" database.
7. Try to load the sample app b in a test store, and an error will be thrown because it will try to query the "shop_session" table in "sample-app-b" database even though it does not exist.

treboryx:

Fixes #585 

Checks if the `shopify_sessions` table exists based on the table name provided. Previously it was just checking within  `information_schema.tables` if the table name exists, which could lead to an issue for hosting multiple apps in the same host.

## Type of change

## Checklist


src/auth/session/storage/mysql.ts
```diff
@@ -121,9 +121,12 @@ export class MySQLSessionStorage implements SessionStorage {
 
   private async hasSessionTable(): Promise<boolean> {
     const query = `
-      SELECT * FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_NAME = ?;
+      SELECT * FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_NAME = ? AND TABLE_SCHEMA = ?;
```
cquemin, line 124: Since we only care about the number of rows that this request returns, I would suggest to use a column for the query, eg. `TABLE_NAME` so that we are more likely to get the value from an index'ed column and therefore make the request a bit more efficient.

So the request would become `SELECT TABLE_NAME FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_NAME = ? AND TABLE_SCHEMA = ?;`

cquemin (changes requested):
Worth considering querying a named column as opposed to all of them to leverage potential usage from an index on said column and other optimization made speciall on this table.

src/auth/session/storage/mysql.ts
treboryx: Good idea!

cquemin:
Did you run it locally against an example app to ensure everything works as expected ?

cquemin:
Also, worth noting that we are changing how session are managed in v6 which should be released in a couple of weeks max. I will check if we need to cherry pick this change as well. Because I suspect this may be lost if you merge it on main or take a long time before v5 becomes deprecated and therefore v6 gets merged on main.

cquemin:
Thank you for contributing to the ongoing improvement of this library - we really appreciate it. Yes you are right, I assume you were part of Shopify. Sorry, that’s my bad.
In order to better support this particular piece of functionality, we have moved it to its own repo.  As a result, we’re closing this PR because the underlying source has been moved to a dedicated new repo:`[ shopfi-app-js](https://github.com/Shopify/shopify-app-js)`.

Since we have changed how session is handled in v6 of `shopify-api-node `and we also added a battery of unit tests, I have reported your changes [there](https://github.com/Shopify/shopify-app-js/pull/34) and updated the unit tests as well.

Thank you!

PR #610, opened 2022-11-25
Update docs with references to session storage packages

mkevinosullivan:

Creates references to the session storage packages that have been created.

### DIFF
--- a/docs/example-migration-v5-node-template-to-v6.md
+++ b/docs/example-migration-v5-node-template-to-v6.md
@@ -34,7 +34,7 @@ pnpm install
 
 ### 4. Create a new file called `shopify.js`
 
-This will be used to to create the shopify api instance used throughout the application.  The `Shopify.Context` configuration as well as the example billing configuration that is currently in `index.js` will also be moved into this file.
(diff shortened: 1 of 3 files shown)


teddyhwang (approved):

docs/example-migration-v5-node-template-to-v6.md
```diff
@@ -135,53 +135,35 @@ This file needs to be updated to use the `shopify` api instance as well as the n
  }
 ```
 
-### 6. Copy the `SQLiteSessionStorage` class from the `@shopify/shopify-app-express` package
+### 6. Install the `@shopify/shopify-app-session-storage-sqlite` package
```
cquemin, line 138: This is not for this line but in Point 4, you mention `Shopify.Context` but for v6, this is replaced by `Shopify.config` if I made no mistake ?

cquemin (changes requested):
I believe we still reference wrongly `Shopify.Context` in the readme.

docs/example-migration-v5-node-template-to-v6.md
mkevinosullivan: I've left it as-is (I added "v5" just before the mention to help clarify), because the instruction is to take the configuration in the current template, the current v5 `Shopify.Context` configuration, and move it into the `shopify.js` file being created.  I think I still need to refer to `Shopify.Context` as that's the migration part.

PR #619, opened 2022-12-05
Add deprecation logger method

paulomarg:

In order to make it easier for developers to spot upcoming breaking changes, we're introducing a new `deprecate` method that will trigger a warning-level log.

Adding that method, and adding a check to make this method break once we reach a certain version of the library, to prevent us from going over the desired release version that fully removes the deprecated code without actually doing it.

The check only applies if `process.env.CI` is true.

## Checklist

### DIFF
--- a/lib/logger/__tests__/logger.test.ts
+++ b/lib/logger/__tests__/logger.test.ts
@@ -1,4 +1,6 @@
+import {FeatureDeprecatedError} from '../../error';
 import {LogSeverity} from '../../types';
+import {SHOPIFY_API_LIBRARY_VERSION} from '../../version';
 import {shopify} from '../../__tests__/test-helper';
 
 describe('shopify.logger', () => {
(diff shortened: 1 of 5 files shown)


lib/logger/__tests__/logger.test.ts
```diff
@@ -80,6 +94,33 @@ describe('shopify.logger', () => {
...
+        testContext: true,
+      });
+
+      expect(shopify.config.logger.log).toHaveBeenCalledWith(
+        LogSeverity.Warning,
+        expect.stringContaining('[Deprecation warning]'),
+      );
+    });
+
+    it('fails to log deprecations for older versions on CI', async () => {
```
nelsonwittwer, line 113: If it is past the deprecation schedule, shouldn't the functions be removed from the library?

I'd think the failing assertions would be in context of the functions that are deprecated.

```js
function coolFunctionBro() {
  shopify.logger("2.0.0")
}

# test
test("coolFunctionBro", () => {
  #expect shopify.logger.deprecate called
  # compare within deprecation schedule as a separate assertion
});
```

mkevinosullivan (approved):

Issue #625, opened 2022-12-07
v6: InvalidWebhookError: No HTTP webhooks registered for topic products/update


mariusa:
# Issue summary

Trying to migrate from v5 to v6. Updated code as per guide.

Webhooks are registered for topics such as `PRODUCTS_UPDATE`
but `shopify.webhooks.process()` gives
`InvalidWebhookError: No HTTP webhooks registered for topic products/update`

Looks like `products/update` got forgotten to be handled as `PRODUCTS_UPDATE` internally?

If we do the registration wrong, at least the error message should reference to correct topic.

## Checklist

- [x] I have described this issue in a way that is actionable (if possible)

paulomarg:
Hi again! This error is thrown when the library can't find the handler to call for a specific topic, as you mentioned. Just to make sure the set up is correct here, are you adding a handler for that topic, as per https://github.com/Shopify/shopify-api-js/blob/main/docs/migrating-to-v6.md#changes-to-webhook-functions?

I assume your call looks like this?

```ts
await shopify.webhooks.addHandlers({
  PRODUCTS_UPDATE: {
    deliveryMethod: DeliveryMethod.Http,
    callbackUrl: shopify.config.webhooks.path,
    callback: () => { /* ... */ },
  },
});
```

> If we do the registration wrong, at least the error message should reference to correct topic.

That makes sense. `products/update` is the format Shopify passes the topic in, but we do translate it to `PRODUCTS_UPDATE` internally to match the GraphQL API. We should log the translated topic here.

mariusa:
Hi & thanks,

> are you adding a handler for that topic,

Yes, when the app starts, similar to your code example
`shopify.config.webhooks.path` doesn't exist, so using

```js
let handlers2 = {
        PRODUCTS_UPDATE: {
            deliveryMethod: 'http',
            callbackUrl: `https://pastebin.com`,
            callback: () => { console.log('test PRODUCTS_UPDATE') },
        }
    }

    console.log('handlers2', handlers2)
    await shopify.webhooks.addHandlers(handlers2)
    const topicsAdded = shopify.webhooks.getTopicsAdded()
    console.log('topicsAdded', topicsAdded)
```

gives
```
handlers2 {
  PRODUCTS_UPDATE: {
    deliveryMethod: 'http',
    callbackUrl: 'https://pastebin.com',
    callback: [Function: callback]
  }
}
topicsAdded [ 'PRODUCTS_UPDATE' ]
```

[...]

PR #626, opened 2022-12-07
Log translated webhook topic when not found

paulomarg:

Part of #625 

Currently, we're using `topic` when throwing an error because a webhook handler was not found. While it isn't incorrect, we should be using `graphqlTopic` to show the `PRODUCTS_UPDATE` format rather than` products/update`, so that the error isn't misleading for callers.

Changing the variable we use for the error message to make the message consistent with the configuration.

## Type of change

## Checklist

### DIFF
--- a/lib/webhooks/__tests__/utils.ts
+++ b/lib/webhooks/__tests__/utils.ts
@@ -25,7 +25,7 @@ export function headers({
   apiVersion = '2023-01',
   domain = 'shop1.myshopify.io',
   hmac = 'fake',
-  topic = 'PRODUCTS_CREATE',
+  topic = 'products/create',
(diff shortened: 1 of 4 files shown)


lib/webhooks/__tests__/utils.ts
```diff
@@ -23,7 +23,7 @@ export function getTestExpressApp() {
 
 export function headers({
   hmac = 'fake',
-  topic = 'PRODUCTS_CREATE',
+  topic = 'products/create',
```
cquemin, line 26: In the description of the PR you are saying : "we should be using graphqlTopic to show the PRODUCTS_UPDATE format rather than products/update", but here it seems that you're doing the opposite??

PR #627, opened 2022-12-07
Update graphql client docs

nickwesselman:
This seems to be the appropriate usage of the GraphQL client now.


paulomarg (approved):
Thanks, great catch!

PR #630, opened 2022-12-08
Add processing of X-Shopify-API-Version to webhook processing

mkevinosullivan:

API library component for https://github.com/Shopify/shopify-app-js/issues/29

Add processing of `X-Shopify-API-Version` header to webhook processing, and pass it as an optional parameter to the webhook handler.

## Type of change

## Checklist

### DIFF
--- a/lib/webhooks/types.ts
+++ b/lib/webhooks/types.ts
@@ -12,7 +12,7 @@ export type WebhookHandlerFunction = (
   shop_domain: string,
   body: string,
   webhookId: string,
-  apiVersion: string,
+  apiVersion?: string,
 ) => Promise<void>;
 
 interface BaseWebhookHandler {


lib/webhooks/process.ts
```diff
@@ -68,7 +69,13 @@ export function process(
           log.debug('Found HTTP handler, triggering it', loggingContext);
 
           try {
-            await handler.callback(graphqlTopic, domain, rawBody, webhookId);
+            await handler.callback(
```
paulomarg, line 72: The number of arguments here is getting a little bit out of hand, it'd be nice to make this an object in a future breaking change.

lib/webhooks/types.ts
```diff
@@ -12,6 +12,7 @@ export type WebhookHandlerFunction = (
   shop_domain: string,
   body: string,
   webhookId: string,
+  apiVersion: string,
```
paulomarg, line 15: We should make this optional (just for the type) so this isn't a breaking change for TS apps.

paulomarg (approved):
Just one nit, but LGTM otherwise.

lib/webhooks/process.ts
mkevinosullivan: I was thinking the same thing while making the change!

PR #635, opened 2022-12-13
'Load a session from storage' typo

scottdixon:
Hey @paulomarg!

We just discussed - replaces `await` with `new` in the **Load a session from storage** docs.


paulomarg (approved):
🔥 thanks!

PR #639, opened 2022-12-15
Export deprecation notices for v5 main interface

paulomarg:

Fixes https://github.com/Shopify/first-party-library-planning/issues/517

The differences between v5 and v6 of this library are very extensive. In order to meet developers closer to where they are when migrating, we should make it easy for them to find the migration instructions directly from their IDE.

Exporting an interface that replicates the previous version's main exports, which point the user toward the appropriate section of the migration guide.

## Type of change

## Checklist

cc @developit

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -1,6 +1,7 @@
 import {loadRestResources} from '../rest/load-rest-resources';
 import {ShopifyRestResources} from '../rest/types';
 
+import {DeprecatedV5Types} from './deprecated-v5-types';
 import {ConfigParams, ConfigInterface} from './base-types';
 import {validateConfig} from './config';
 import {clientClasses, ShopifyClients} from './clients';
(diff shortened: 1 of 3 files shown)


lib/index.ts
```diff
@@ -21,6 +22,10 @@ export * from './session/types';
 export * from './auth/types';
 export * from './webhooks/types';
 
+// Temporarily export the deprecated v5 types as a Shopify object (as opposed to the type above) to help folks find
+// the migration guide.
+export const Shopify: DeprecatedV5Types = {};
```
paulomarg, line 27: This slightly conflicts with the _type_ `Shopify` we export above, but TS seems to be capable of telling them apart when importing the value.

cquemin (approved):
LGTM

lib/index.ts
developit: couldn't this be merged into the Shopify interface?

```
export interface Shopify<
  T extends ShopifyRestResources = ShopifyRestResources,
> extends DeprecatedV5Types {
  // ...
}
```

PR #644, opened 2022-12-20
Fix AssignedFulfillmentOrder body parsing

### LINKED ISSUE #368: AssignedFulfillmentOrder.all() Always Returns an Empty Array
# Issue summary

The `AssignedFulfillmentOrder.all` method is returning an empty array even though there are fulfillment orders assigned to the app that is making the call. I have verified that the following two examples are producing different results given the same inputs using API version 2022-04.

```javascript
const session = await Shopify.Utils.loadOfflineSession(shop);
const fulfillmentOrders = await AssignedFulfillmentOrder.all({ session, assignment_status: 'fulfillment_requested' });
// fulfillmentOrders is empty
```

```javascript
const session = await Shopify.Utils.loadOfflineSession(shop);
const client = new Shopify.Clients.Rest(session.shop, session.accessToken);
const response = await client.get({
  path: 'assigned_fulfillment_orders',
  query: { assignment_status: 'fulfillment_requested' }
});
// response.body.fulfillment_orders is not empty
```

The issue is that neither the `NAME` or `PLURAL_NAME` fields of the `AssignedFulfillmentOrder` class match the 'fulfillment_orders' field in the response body. The applicable code can be found in the `Base.createInstancesFromResponse()` method [here](https://github.com/Shopify/shopify-node-api/blob/9297dda7da2c266b9987d65cba693abbe7ab176a/src/base-rest-resource.ts#L210). To fix this you could just change the `NAME` and `PLURAL_NAME` fields to 'fulfillment_order' and 'fulfillment_orders' respectively.

## Expected behavior

The method should return the matching resources.

## Actual behavior

The method returns an empty array.

## Checklist

### LINKED ISSUE #642: Assigned Fulfillment Orders
# Issue summary

The AssignedFulfillmentOrder.all method is returning an empty array even though there are fulfillment orders assigned to the app that is making the call. I have verified that the following two examples are producing different results given the same inputs using API version 2022-10. ↓

const session = await Shopify.Utils.loadOfflineSession(shop);
const fulfillmentOrders = await AssignedFulfillmentOrder.all({ session, assignment_status: 'fulfillment_requested' });
// fulfillmentOrders is empty
## Expected behavior

The method should return the matching resources

 ##  Actual behavior

The method returns an empty array.

## Checklist

paulomarg:

Fixes #368 
Fixes #642 

There was a bug in the `AssignedFulfillmentOrder` class where the `NAME` and `PLURAL_NAME` didn't match the actual data sent by the API.

I've updated our auto-generation code to allow overriding these values, so I'm updating the auto-generated resources here.

## Type of change

## Checklist


cquemin (approved):
I can see a lot of duplication in the test data. Would there be a way to put the test data in common ?

mkevinosullivan (approved):

PR #648, opened 2022-12-21
Export missing client types from package

### LINKED ISSUE #646: Better Type Exports for Http Clients
## Overview/summary

During development of a typescript based application using the v6 of this package, I have faced an issue with the supported type exports configured in the package. As of my understanding it would be way more convenient to have access to the defined types in `lib/clients/types` to e.g. use the predefined enums for the request types, which are necessary for a rest client request.

Example code what I would like to achieve:

```js
	import { DataType } from '@shopify/shopify-api/lib.clients/types // <=== import it from here

	const data = await restClient.put({
		path: `customers/123456789/addresses/123456789`,
		type: DataType.JSON, // <=== use it here
		data: {...}
	})
```

Unfortunately, the import is not possible due to limitations inside of the package.json
As of my understanding we either have to add the path to the package.json or chain the export til an available export path is reached. Happy to discuss that with you! 😊 

## Motivation

I am actively using this package and would like to contribute to it / support the community. 

> What inspired this enhancement?

Daily usage of the package

### Area

- [ ] Dunno what to add here
---

## Checklist

- [X] I have described this enhancement in a way that is actionable (if possible)

paulomarg:

Fixes #646 

We were missing an export for the client types in `src/index.ts`.

Exporting the missing types.

## Type of change

## Checklist


cquemin (approved):
Nice and easy fix!

PR #649, opened 2022-12-21
Add SDK headers to SF API client

paulomarg:

The Storefront API has been updated to allow a few headers to improve how we track requests to it, and our client should start complying with it.

Making the override method where we set the headers for the GraphQL client more generic, so that it takes any number of headers, and setting the new headers in the SF API client.

## Type of change

Considering this a patch from the perspective of it being a fix for tracking, rather than adding new features for users of this library.

## Checklist

### DIFF
--- a/lib/clients/graphql/storefront_client.ts
+++ b/lib/clients/graphql/storefront_client.ts
@@ -1,5 +1,5 @@
 import {SHOPIFY_API_LIBRARY_VERSION} from '../../version';
-import {ShopifyHeader} from '../../types';
+import {LIBRARY_NAME, ShopifyHeader} from '../../types';
 import {httpClientClass} from '../http_client/http_client';
 import {Session} from '../../session/session';
 import {HeaderParams} from '../http_client/types';
@@ -32,7 +32,7 @@ export class StorefrontClient extends GraphqlClient {
         ? this.storefrontClass().CONFIG.privateAppStorefrontAccessToken ||
           this.storefrontAccessToken
         : this.storefrontAccessToken,
-      [ShopifyHeader.StorefrontSDKVariant]: '@shopify/shopify-api',
+      [ShopifyHeader.StorefrontSDKVariant]: LIBRARY_NAME,
       [ShopifyHeader.StorefrontSDKVersion]: SHOPIFY_API_LIBRARY_VERSION,
     };
   }
--- a/lib/clients/graphql/storefront_client.ts
+++ b/lib/clients/graphql/storefront_client.ts
@@ -26,13 +26,15 @@ export class StorefrontClient extends GraphqlClient {
   }
 
   protected getApiHeaders(): HeaderParams {
+    const sdkVariant = LIBRARY_NAME.toLowerCase().split(' ').join('-');
+
     return {
       [ShopifyHeader.StorefrontAccessToken]: this.storefrontClass().CONFIG
         .isPrivateApp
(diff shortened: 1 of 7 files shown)


lib/clients/graphql/storefront_client.ts
```diff
@@ -27,13 +25,15 @@ export class StorefrontClient extends GraphqlClient {
...
     return {
-      header: ShopifyHeader.StorefrontAccessToken,
-      value: (this.storefrontClass().CONFIG.isPrivateApp
+      [ShopifyHeader.StorefrontAccessToken]: this.storefrontClass().CONFIG
+        .isPrivateApp
         ? this.storefrontClass().CONFIG.privateAppStorefrontAccessToken ||
           this.storefrontAccessToken
-        : this.storefrontAccessToken) as string,
+        : this.storefrontAccessToken,
+      [ShopifyHeader.StorefrontSDKVariant]: '@shopify/shopify-api',
```
cquemin, line 35: Should this also be a constant that lives next to `SHOPIFY_API_LIBRARY_VERSION` ?

lib/clients/http_client/types.ts
```diff
@@ -6,11 +6,13 @@ export interface HeaderParams {
   [key: string]: string | number | string[];
 }
 
+/* eslint-disable @shopify/typescript/prefer-pascal-case-enums */
 export enum DataType {
-  JSON = 'application/json', // eslint-disable-line @shopify/typescript/prefer-pascal-case-enums
-  GraphQL = 'application/graphql', // eslint-disable-line @shopify/typescript/prefer-pascal-case-enums
-  URLEncoded = 'application/x-www-form-urlencoded', // eslint-disable-line @shopify/typescript/prefer-pascal-case-enums
+  JSON = 'application/json',
```
cquemin, line 11: Are they no Javascript constant we could use instead of defining them ourselves? I know it would not add much value but I am just cautious redefining things manually as it always open the door to a future mistake

cquemin (commented):
Few minor questions regarding constant definition.

lib/clients/graphql/storefront_client.ts
paulomarg: That's a good point. I wouldn't put it along with the version, but we should use the same value we use in our User-Agent headers. I'll centralize this.

lib/clients/http_client/types.ts
paulomarg: Unfortunately no, we use `@shopify/network` for those shared constants. It has pre-defined values for the header names (which we use here), but not for the MIME types :(

cquemin: Ah ok I understand worth raising an issue there or even a PR no ? Or do you think thats not the right place for this sort of constant?

lib/types.ts
```diff
@@ -13,17 +13,22 @@ export enum ApiVersion {
   Unstable = 'unstable',
 }
 
+export const LIBRARY_NAME = 'Shopify API Library';
```
melissaluu, line 16: For the `X-SDK-Variant` header, we'd prefer a value that's lower case and hyphenated instead of using spaces. Since this `LIBRARY_NAME` is used in the user agent value, maybe we can do something like this while setting the variant header or create a new const:

```
[ShopifyHeader.StorefrontSDKVariant]: LIBRARY_NAME.toLowerCase().split(' ').join('-'),
```

paulomarg: I can do that!

PR #650, opened 2022-12-21
Add debug log of version and environment at setup

mkevinosullivan:

Logging the library version and the runtime environment string at initialization would help with debugging of issues.

Add a call to `shopify.logger.debug` with the above information to `shopifyApi`

## Type of change

## Checklist

- [ ] ~I have documented new APIs/updated the documentation for modified APIs (for public APIs)~ - not applicable

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -1,5 +1,6 @@
 import {loadRestResources} from '../rest/load-rest-resources';
 import {ShopifyRestResources} from '../rest/types';
+import {abstractRuntimeString} from '../runtime/platform';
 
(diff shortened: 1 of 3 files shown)


lib/index.ts
```diff
@@ -67,5 +69,11 @@ export function shopifyApi<T extends ShopifyRestResources>(
     }) as T;
   }
 
+  shopify.logger
+    .debug(
```
paulomarg, line 73: I was thinking, this is a fairly useful thing to log as it also serves the purpose of letting the app know we're ready for it. Should this be an info log since it only happens once?

paulomarg (approved):

PR #652, opened 2022-12-22
Don't ignore previous headers when beginning OAuth

### LINKED ISSUE #651: Enable a way to pass information through the OAuth Process
## Overview

Switching to the new v6 of this library we faced a problem with shopifys admin links.

We have to use several admin links for our application that e.g. link the merchant from his order list straight to our app including the selected order ids. If we run an app using the online sessions we have to reauthenticate once a day, cause the session expires which results in the first admin links used forwarding the merchant back to the root page of our app instead of opening the right page. To solve this, we will need to pass the referrer / initially requested url through the OAuth process to redirect the user after successfully passing the process back to this url. The new shopify.auth.callback / shopify.auth.begin functions do not support to pass any additional data and the state variable is used for the nonce as of security reasons (so far my understanding). Further adjusting the state variable or cookie content is not possible due to the fact, that this library takes care of the whole implementation.

Headers are also removed for a node.js / typescript based application - so no way to add the additional information here.

Thus we end up in the situation, that we either have to fallback to offline sessions or in the described scenario, where the user gets redirected to a static route in our app once successfully completed the flow - which is not ideal tbh.

Is there a way to get around this problem without switching to offline sessions that I have missed?

If not I would like to request it as a new feature.

paulomarg:

Fixes #651 

As pointed out in the issue, there is currently no way of adding custom headers to a response when beginning OAuth on Node.js environments, because we overwrite all headers with the ones we need to set. We need to preserve any headers previously set by the app when adding our own.

Changing the oauth begin process to start out from the incoming response via a new conversion method (defaulting to an empty one for non-node environments), and then adding our cookies / redirect headers on top of those rather ignoring all of them.

## Type of change

## Checklist


cquemin (approved):
Nice fix!

Issue #427, opened 2022-07-21
Webhook registry design issue


richardscarrott:
# Issue summary

I'm working on migrating an old Shopify app to the new [@shopify/app](https://www.npmjs.com/package/@shopify/app) node template.

I've got a few concerns with the webhook "registry" API design.

1. Given the recommended usage whereby you call `registerAll` in the oauth callback, _new_ or _removed_ webhooks may never get registered / reregistered.
    - If using offline tokens, there's every chance your app offers very little functionality in it's UI (e.g. a sales channel) meaning users may never, or at least very rarely go back through the auth flow so any new or removed webhooks will never get registered / reregistered.
1. It only allows a single webhook per topic
    - We have multiple services which listen to various webhooks (e.g. products service and orders service); if two services were interested in the same topic (e.g. app/uninstalled), [it looks like this library would break the previous webhook](https://github.com/Shopify/shopify-api-node/blob/main/src/webhooks/registry.ts#L165)?
    - If this is the case, it seems very subtle and not in-keeping with the underlying Shopify Webhook API?

## Expected behavior

Just a suggestion but a more practical and idiomatic nodejs / express API might be to
a) separate the route handler from registering of webhooks and
b) use middleware

```ts
const withShopifyWebhook = (
  req: Request,
  res: Response,
  next: NextFunction
) => {
  const topic = req.get('X-Shopify-Topic');
  const shop = req.get('X-Shopify-Shop-Domain');
  req.shopifyWebhook = { topic, shop };
  next();
};

const validateShopifyWebhook = (
  req: Request,
  res: Response,
  next: NextFunction
) => {
  assert(req.shopifyWebhook.topic);
  assert(req.shopifyWebhook.shop);
  // TODO: hmac validation
  next();
};

app.post('/api/webhooks/', withShopifyWebhook, validateShopifyWebhook, bodyParser, async (req, res) => {
   // Process webhook here, e.g.
   await webhookQueue.push({ shop: req.shopifyWebhook.shop, topic: req.shopifyWebhook.topic,  body: req.body });
   res.sendStatus(204);
})
```

Then for registry, offer up a util which registers a unique webhook by `topic` and `address`:

```ts
const ensureShopifyWebhook = async (shop: string, accessToken: string, topic: string, address: string) => {
    const existingWebhook = await shopify.webhook.find(topic, address);
    if (existingWebhook) {
       return;
    }
    await shopify.webhook.create({ topic, address });
}

[...]

PR #654, opened 2022-12-23
Export function that only validates webhooks

paulomarg:

We addressed most of the suggestions in #427, but I realized there was one more thing we could do here: right now, when processing webhooks, we only export a function that takes over and does everything for the app.

Attempting to make it easier for apps to e.g. use a processing queue for webhooks, but still be able to use the library's validation to ensure the requests are coming from Shopify.

I'm breaking out the code to validate a request into a separate `shopify.webhooks.validate` call, and calling through to that from `shopify.webhooks.process` so they behave the same way, but still allow apps to not trigger the callback.

For example, this would now be possible:

```ts
shopify.webhooks.addHandlers({
  APP_UNINSTALLED: {
    deliveryMethod: DeliveryMethod.Http,
    callbackUrl: '/webhooks',
    callback: async (a, b, c, d, e) => console.log(a, b, c, d, e),
  },
});

const app = express();

app.post(
  '/webhooks',
  express.text({ type: "*/*" }),
  async (req, res, next) => {
    try {
      const webhookCheck = await shopify.api.webhooks.validate({
        rawBody: req.body,
        rawRequest: req,
        rawResponse: res,
      });

      if (webhookCheck.valid) {
        console.log(
          `Webhook checks out for topic ${webhookCheck.topic}, queueing for processing`
        );
      } else {
        console.log(
          `Whoops, webhook did not validate: ${JSON.stringify(webhookCheck)}`
        );
      }
    } catch (e) {
      console.log(`Failed to validate webhook: ${e.message}`);
    }

    res.status(200).send("OK");
  }
);
```

which would produce

```
Webhook checks out for topic APP_UNINSTALLED, queueing for processing
...
Whoops, webhook did not validate: {"valid":false,"reason":"missing_headers","missingHeaders":["X-Shopify-API-Version","X-Shopify-Webhook-Id"]}
```

Note: this makes the `callback` field in the webhook config a bit awkward, since it's unnecessary for `validate`-only scenarios. We could consider making it optional and then throwing an error in `process` so apps don't need to create a dummy handler.
Note 2: we might need to reconsider how the express package sets up webhook configs (handlers are currently passed in to the `processWebhooks` middleware, which wouldn't be called here.

## Type of change

## Checklist

I'll document this properly once we figure out this is something we want to do


lib/webhooks/process.ts
```diff
@@ -1,25 +1,43 @@
...
+  WebhookProcessParams,
+  DeliveryMethod,
+  WebhookValidationErrorReason,
+  WebhookValidationMissingHeaders,
+  WebhookValidationValid,
+  WebhookValidationInvalid,
+} from './types';
+import {validate} from './validate';
+
+interface HandlerCallResult {
```
cquemin, line 24: Should `HandlerCallResult` and `ErrorCallResult` be 'merged' into one unique `CallResult` since they have the structure?

cquemin (approved):
One question for me, but not a blocking  one.

lib/webhooks/process.ts
```diff
@@ -1,109 +1,84 @@
...
-import {topicForStorage} from './registry';
+import {
+  WebhookRegistry,
+  WebhookProcessParams,
+  DeliveryMethod,
+  WebhookValidationErrorReason,
+  WebhookValidationMissingHeaders,
+  WebhookValidationValid,
+  WebhookValidationInvalid,
+  HttpWebhookHandlerWithCallback,
```
mkevinosullivan, line 21: Again, complete nit: alphabetical order for imports

lib/webhooks/types.ts
```diff
@@ -115,3 +120,41 @@ export interface AddHandlersParams {
 export interface WebhookProcessParams extends AdapterArgs {
   rawBody: string;
 }
+
+export interface WebhookValidateParams extends AdapterArgs {
+  rawBody: string;
+}
```
mkevinosullivan, line 126: Any reason not to do this?
```suggestion
export interface WebhookValidateParams extends WebhookProcessParams {}
```

lib/webhooks/validate.ts
```diff
@@ -0,0 +1,124 @@
...
+  [property in keyof WebhookFields]: ShopifyHeader;
+} = {
+  apiVersion: ShopifyHeader.ApiVersion,
+  domain: ShopifyHeader.Domain,
+  hmac: ShopifyHeader.Hmac,
+  topic: ShopifyHeader.Topic,
+  webhookId: ShopifyHeader.WebhookId,
+};
+
+export function validate(config: ConfigInterface) {
```
mkevinosullivan, line 32: Should we start calling these _factories_ like we do in the _other_ project 😉 , e.g., 
```suggestion
export function validateFactory(config: ConfigInterface) {
```

mkevinosullivan (approved):
Just nits

PR #655, opened 2022-12-23
Add response headers to GraphqlQueryError

### LINKED ISSUE #553: Headers aren't included in GraphQL error (maybe REST too)
# Issue summary

Hi, the headers are not included in response when there's an error:

```js
let products;
try {
  products = await client.query({
    data: `{
        products (first: 10) {
          edges {
            node {
              id
              title
              descriptionNotExistentToGetERror
            }
          }
        }
      }`,
  });
} catch (error) {
  if (error instanceof ShopifyErrors.GraphqlQueryError) {
// goes here
   //error.response.errors is set
// err.response.headers  is not set
  } else {
    // handle other errors
  }
  // the entire err object doesn't include any headers
}
```

## Expected behavior

err should include response.headers, but it's not included. Shopify code:
```js
 throw new ShopifyErrors.GraphqlQueryError({
                                message: 'GraphQL query returned errors',
                                response: result.body,
                            });
```

---

## Checklist

See also https://github.com/Shopify/shopify-api-node/pull/340

cquemin:

Fixes #553 (https://github.com/Shopify/shopify-api-js/issues/553)

When a `GraphqlQueryError` is raised we were not adding in the error the response's headers. Those can be useful to understand the error. So this is now rectified

## Type of change

## Checklist

### DIFF
--- a/lib/clients/graphql/__tests__/graphql_client.test.ts
+++ b/lib/clients/graphql/__tests__/graphql_client.test.ts
@@ -195,6 +195,11 @@ describe('GraphQL client', () => {
       }`,
     };
 
+    const expectedHeaders = {
+      'Content-Type': 'application/graphql',
+      'X-Shopify-Access-Token': 'any_access_token',
+    };
+
     const errorResponse = {
       data: null,
       errors: [
@@ -220,6 +225,7 @@ describe('GraphQL client', () => {
           },
         },
       },
+      headers: expectedHeaders,
     };
 
     queueMockResponse(JSON.stringify(errorResponse));
@@ -228,6 +234,7 @@ describe('GraphQL client', () => {
       new ShopifyErrors.GraphqlQueryError({
         message: 'GraphQL query returned errors',
         response: errorResponse,
+        headers: expectedHeaders,
       }),
     );
 
--- a/lib/clients/graphql/graphql_client.ts
+++ b/lib/clients/graphql/graphql_client.ts
@@ -82,6 +82,7 @@ export class GraphqlClient {
       throw new ShopifyErrors.GraphqlQueryError({
         message: 'GraphQL query returned errors',
         response: result.body as unknown as {[key: string]: unknown},
+        headers: result.headers,
       });
     }
     return result;
--- a/.changeset/stupid-roses-behave.md
+++ b/.changeset/stupid-roses-behave.md
@@ -1,0 +1,5 @@
+---
+'@shopify/shopify-api': patch
+---
+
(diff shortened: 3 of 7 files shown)


lib/clients/graphql/__tests__/graphql_client.test.ts
```diff
@@ -223,6 +229,7 @@ describe('GraphQL client', () => {
       new ShopifyErrors.GraphqlQueryError({
         message: 'GraphQL query returned errors',
         response: errorResponse,
+        headers: expectedHeaders,
```
mkevinosullivan, line 232: If `errorResponse` contains a `headers` property, would it just be available at `response.headers`?  Why add a separate `headers` property to the `GraphqlQueryError`?

lib/clients/graphql/graphql_client.ts
```diff
@@ -79,6 +79,7 @@ export class GraphqlClient {
       throw new ShopifyErrors.GraphqlQueryError({
         message: 'GraphQL query returned errors',
         response: result.body as unknown as {[key: string]: unknown},
+        headers: result.headers,
```
mkevinosullivan, line 82: I wonder if the idiomatic expectation would be that `GraphqlQueryError` contains a `response` property that itself then contains a `body` property and a `headers` property (in alignment with a Node `response` object).

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 - [Patch] Don't ignore previous headers when beginning OAuth [#652](https://github.com/Shopify/shopify-api-js/pull/652)
 - [Patch] Export missing client types from package [#648](https://github.com/Shopify/shopify-api-js/pull/648)
 - [Patch] Add an info-level log of API library version and runtime environment string during initialization, to aid in troubleshooting [650](https://github.com/Shopify/shopify-api-js/pull/650)
+- [Patch] Add response headers to `GraphqlQuerryError` [#655](https://github.com/Shopify/shopify-api-js/pull/655)
```
mkevinosullivan, line 13:
```suggestion
- [Patch] Add response headers to `GraphqlQueryError` [#655](https://github.com/Shopify/shopify-api-js/pull/655)
```

lib/clients/graphql/graphql_client.ts
mkevinosullivan: Let's leave your change as-is.  Setting `response` to contain `body` and `headers` would become an unnecessary breaking change, I think.

mkevinosullivan (approved):

PR #671, opened 2023-01-06
Add Hmac timestamp validation on OAuth

### LINKED ISSUE #570: Validating the timestamp value during oauth
When OAuth occurs, the server must validate that the HMAC matches.

That's fine, but what about the timestamp? Shouldn't there be some validation to check that the timestamp parameter is within some certain tolerance threshold to avoid someone trying to reuse a valid HMAC (found in some logs for example)?

cquemin:

Fixes #570 https://github.com/Shopify/shopify-api-js/issues/570

During the OAuth flow we receive an HMAC that has been computer over URL parameters. One of them is a timestamp. This timestamp wasn't validated therefore allowing potential replay attack to occur.

This PR add timestamp validation to ensure the submitted timestamp is within a reasonable windows in the past or in the future
The tolerance should be big enough that it doesn't break anything. However if servers are widely desynchronized then this may be an issue, arguably something that 3p should fix. But does it means this should be a breaking change on our end?

## Type of change

## Checklist

### DIFF
--- a/lib/utils/__tests__/hmac-validator.test.ts
+++ b/lib/utils/__tests__/hmac-validator.test.ts
@@ -101,3 +101,29 @@ test('hmac with timestamp older than 10 seconds throws InvalidHmacError', async 
     ShopifyErrors.InvalidHmacError,
   );
 });
+
+test('hmac with timestamp more than 10 seconds in the future throws InvalidHmacError', async () => {
+  shopify.config.apiSecretKey = 'my super secret key';
+  const submittedTimestamp = String(Date.now() + 11000);
+  const queryString = `code=some+code+goes+here&shop=the+shop+URL&state=some+nonce+passed+from+auth&timestamp=${submittedTimestamp}`;
+  const queryObjectWithoutHmac = {
+    code: 'some code goes here',
+    shop: 'the shop URL',
+    state: 'some nonce passed from auth',
+    timestamp: submittedTimestamp,
+  };
+
+  const localHmac = crypto
+    .createHmac('sha256', shopify.config.apiSecretKey)
+    .update(queryString)
+    .digest('hex');
+
+  const testQuery: AuthQuery = Object.assign(queryObjectWithoutHmac, {
+    hmac: localHmac,
+  });
+
+  const validateHmac = shopify.utils.validateHmac;
+  await expect(validateHmac(testQuery)).rejects.toBeInstanceOf(
+    ShopifyErrors.InvalidHmacError,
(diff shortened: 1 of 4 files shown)


lib/utils/hmac-validator.ts
```diff
@@ -7,6 +7,8 @@ import {safeCompare} from '../auth/oauth/safe-compare';
 
 import ProcessedQuery from './processed-query';
 
+const HMAC_TIMESTAMP_PERMITTED_CLOCK_TOLERANCE_SEC = 10;
```
paulomarg, line 10: Also, (sorry) are we sure 10s is enough for this, with slow networks and such? I think we could be a little more lenient, maybe a minute or two?

cquemin: I wasn't too sure, I used the same value as we used for Json validation. But in this case we may have a processing queue that is slow you are right. I'll put it to 90sec then Hopefully that should be enough. 
I am also conscious hat too much leniency leaves room for more vulnerability...

paulomarg: Agreed on all counts, I think 90s is a reasonable compromise. 60 would probably be enough.

lib/utils/__tests__/hmac-validator.test.ts
```diff
@@ -75,3 +76,55 @@ test('queries with extra keys are included in hmac querystring', async () => {
...
+    hmac: localHmac,
+  });
+
+  const validateHmac = shopify.utils.validateHmac;
+  await expect(validateHmac(testQuery)).rejects.toBeInstanceOf(
+    ShopifyErrors.InvalidHmacError,
+  );
+});
+
+test('hmac with timestamp more than 10 seconds in the future throws InvalidHmacError', async () => {
```
paulomarg, line 106: Nit: it's no longer 10 seconds :)

lib/utils/hmac-validator.ts
```diff
@@ -31,9 +33,27 @@ export function validateHmac(config: ConfigInterface) {
...
   };
 }
+
+export function getCurrentTimeInSec() {
+  return Math.trunc(Date.now() / 1000);
+}
+
+function validateHmacTimestamp(query: AuthQuery) {
+  if (
+    Math.abs(getCurrentTimeInSec() - Number(query.timestamp)) >
```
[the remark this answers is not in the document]
paulomarg: I don't think there's any harm in making it more flexible, and I suppose it might end up being in the future.

PR #690, opened 2023-01-24
Make session private for Base class, add getter

mkevinosullivan:

Fixes Shopify/first-party-library-planning#519

To prevent `session` leaking via the REST resource objects, change `this.session` to `this.#session` and add a getter to allow `this.session` to continue to work within the derived resource classes.

## Type of change

## Checklist


tsconfig.json
```diff
@@ -6,7 +6,8 @@
     "declarationMap": true,
     "pretty": true,
     "moduleResolution": "node",
-    "target": "es5",
+    "module": "commonjs",
+    "target": "es2015",
```
mkevinosullivan, line 10: Needed for `tsc` to output CommonJS module format...
```
    "module": "commonjs",
```

Needed for JavaScript private properties (e.g., #session)...
```
    "target": "es2015",
```

rest/base.ts
```diff
@@ -32,6 +37,11 @@ interface GetPathArgs {
   entity?: Base | null;
 }
 
+interface SetClassPropertiesArgs {
+  CLIENT: typeof RestClient;
+  CONFIG: ConfigInterface;
+}
+
 export class Base {
```
paulomarg, line 45: Should we also proxy `toJSON` through `serialize` for consistency?

rest/load-rest-resources.ts
```diff
@@ -24,9 +25,26 @@ export function loadRestResources({
...
+          CONFIG: config,
+        });
+      });
+
+      Object.entries(NewResource.HAS_MANY).map(([_attribute, klass]) => {
+        (klass as typeof Base).setClassProperties({
+          CLIENT: RestClient,
+          CONFIG: config,
+        });
+      });
```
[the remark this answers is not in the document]
paulomarg: Any this also fixes any other requests that depend on a client, like

```ts
product.variants[1].save();
```

which would fail due to the lack of a client before.

tsconfig.json
paulomarg: I don't necessarily know if this will affect any apps importing this package, but support for es2015 should be mostly there for node 14 (which is going EOL soon and is our latest supported version), so I don't expect the target will be a big problem.

paulomarg (approved):

PR #709, opened 2023-02-02
Add guide for creating custom store app

### LINKED ISSUE #700: Set permanent access token for private app installed on only one store.
The [shopifyApi reference](https://github.com/Shopify/shopify-api-js/blob/main/docs/reference/shopifyApi.md) does not have a private app permanent access token property. I have a custom private app that is installed on only one store and I'm trying to understand how to establish a connection using a permanent access token.

- `@shopify/shopify-api` version: 6.1.0
- Node version: v16.16.0
- Operating system: Manjaro Linux

I'm looking at the [Session object](https://github.com/Shopify/shopify-api-js/blob/main/docs/guides/session-storage.md#basics) docs and I'm guesing that a private app connection is made using an **"offline" session**? Based on this assumption, I'm setting a `Session` with `SessionParams`. I'm confused by creating the `shopifyApi` object though to gain access to "offline" sessions, which seems geared to run through oAuth, with the `apiKey` and `apiSecretKey`, and then having to create a `Session` with the permanent access token `accessToken` and this is referred to being "offline". 

**Am I getting close with the code below???** I have yet to test it, but TS and lints are passing.

On the [performing oAuth doc](https://vanguardmil.myshopify.com/admin/settings/apps/development/26860650497/overview), it states "private apps are unable to perform OAuth, because they don't require an access token to interact with API.", but there is no link or documentation from this note, explaining what to do for private apps. And to create an "offline" session, which I think is the way you create a private app connection, it asks for an `accessToken`, which negates the performing oAuth doc note cited, if indeed "offline" sessions are how you create a private app connection.

Adding to the confusion, is why the `shopifyApi` object requires the `apiKey`, `apiSecretKey`, and `scopes` which are not required  for a private app, because private apps do not use oAuth and scopes have already been established when installing the private app in Shopify admin. The `isPrivateApp` flag in the `shopifyApi` object adds to the confusion, as well, because it gives you the impression that you can create a private app connection from the `shopifyApi` object. 

This is a great app, but you shouldn't have to dig around so much to understand and guess how it works for a private app.

```node
// Is this how you initialize a private app connection to gain access to "clients" functionality?
import '@shopify/shopify-api/adapters/node'
import { shopifyApi, LATEST_API_VERSION, Session } from '@shopify/shopify-api'
const debug = process.env.FUNCTIONS_EMULATOR === 'true'

const shopify = shopifyApi({
  apiKey: 'myprivateappkey',
  apiSecretKey: 'myprivateappsecret',
  apiVersion: LATEST_API_VERSION,
  isPrivateApp: true,
  scopes: ['read_products'],
  isEmbeddedApp: false,
  hostName: debug ? '127.0.0.1:5001' : 'shop.myshopify.com',
})

const sessionId = shopify.session.getOfflineId('shop.myshopify.com')
const session = new Session({
  id: sessionId,
  shop: 'shop.myshopify.com,
  state: 'state',
  isOnline: false,
  accessToken: 'permanentAccessToken',
})

const client = new shopify.clients.Rest({ session: session })
```

On older v5 version, the following code, just two lines, gave me access to all REST APIs, but I see this is now [deprecated](https://github.com/Shopify/shopify-api-js/blob/main/docs/migrating-to-v6.md#changes-to-api-clients). If this is just the way it is in v6, I guess creating the `shopifyApi` object, though confusing for a private app, is not a big deal to get access to the `shopify.session.getOfflineId` and `shopify.clients`. IMO, Just needs to be more clear somewhere in the docs what "offline" means, or a better yet a one page doc with the title "How to create a private app connection" that has the exact steps to establish a private app connection, and use the word "private" app rather than "offline".

```node
import Shopify from '@shopify/shopify-api'

const client = new Shopify.Clients.Rest(
  'mystore.myshopify.com',
  'permanentAccessTokenString',
)
```

## Expected behavior

It would be more intuitive if the `shopifyApi` init object gave the option to either enter a private access permanent token or the apiKey and apiSecret. And, if a private access permanent token is provided in the `shopifyApi` object, that the underlying functionality like REST would know that no session was needed. Like in v5, where it took just two lines of code. If this is not in the roadmap, it would be great to have a better example of how to use offline sessions with a private app, and have that documentation on just a single page in the docs for private apps. Covering how to load a Session with a permanent access token, in conjunction with creating the required `shopifyApi` object, and explain why the apiKey, apiSecret, and scopes are required even though you have a permanent access token. 

Finally, on the Shopify doc below, private apps do not use oAuth or sessions, so the definitions and concepts in shopify-api-js are a bit confusing for a private app, IMO.

### [Types of authentication and authorization methods](https://shopify.dev/apps/auth#types-of-authentication-and-authorization-methods)
image

## Actual behavior

Not clear how to use a permanent access token for a custom private app.

1. Create a custom private app.
1. Try to initialize the shopify REST client using a permanent access token.

mkevinosullivan:

Fixes #700 - there's no guide to how to create a custom store app

Adding a guide for ☝🏻!

### DIFF
--- a/docs/guides/custom-store-app.md
+++ b/docs/guides/custom-store-app.md
@@ -1,0 +1,65 @@
+# Creating a store-specific custom app
+
+A custom app is an app that you or a developer builds exclusively for your Shopify store (unlike a public app, which is built to be used by many stores). You can use custom apps to add features to your Shopify admin, access your store's data directly using Shopify's APIs, or extend your online store to other platforms using the Storefront API. Sometimes such a custom app is referred to as a "private app".
+
+## Prerequisites
+
+1. [Enable custom app development from the Shopify admin](https://help.shopify.com/en/manual/apps/custom-apps#enable-custom-app-development-from-the-shopify-admin)
+2. [Create and install a custom app](https://help.shopify.com/en/manual/apps/custom-apps#create-and-install-a-custom-app)
+3. :warning: Have a secure copy of the Admin API access token from the previous step.
+
+## Basic concepts
+
+A store-specific custom app does not use the OAuth process to authenticate - it uses the secrets established during the app creation and install process to access the API.  As a result, there are no sessions to be retrieved from incoming requests and stored in a database, etc.
+
+When initializing `shopifyApi` in a custom app, set the `isPrivateApp` configuration property to `true`, and set the `apiSecretKey` to the **Admin API access token** obtained during the installation process (step 2 in the [prerequisites](#prerequisites)).
+
+The `scopes` configuration property must be set to something for initialization to work.  However, the value is ignored as the scopes of the app are set during the create and install process (step 2 in the [prerequisites](#prerequisites)) and are linked to the Admin API access token.
+
+## Example
+
+### Initialization
+
+```js
+import "@shopify/shopify-api/adapters/node";
+import { shopifyApi, LATEST_API_VERSION, Session } from "@shopify/shopify-api";
+import { restResources } from "@shopify/shopify-api/rest/admin/2023-01";
+
+const shopify = shopifyApi({
+  apiKey: "App_API_key",
+  apiSecretKey: "Admin_API_Access_Token", // Note: this is the API Access Token, NOT the API Secret Key
+  apiVersion: LATEST_API_VERSION,
+  isPrivateApp: true,                     // this MUST be set to true (default is false)
+  scopes: ["read_products"],              // this must have a value but it will be ignored by the library
+  isEmbeddedApp: false,
+  hostName: "my-shop.myshopify.com",
+  // Mount REST resources.
+  restResources,
+});
+```
+
+> **Note** The `apiSecretKey` is **NOT** set to the API secret key but to the **Admin API Access Token**.
+
+### Making requests
+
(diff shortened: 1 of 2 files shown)


docs/guides/custom-store-app.md
```diff
@@ -0,0 +1,65 @@
...
+> **Note** The `apiSecretKey` is **NOT** set to the API secret key but to the **Admin API Access Token**.
+
+### Making requests
+
+API requests, either using `shopify.clients.Rest` or `shopify.clients.Graphql`, or using the REST resources, require a `session` parameter.  Since there are no sessions in a store-specific custom app, a `session` parameter needs to be created.
+
+To create the session object, the `id`, `state` and `isOnline` properties must be populated but will be ignored.  Only the `shop` parameter is required when making the API requests from a store specifc custom app and the value must match the stop on which the custom app is installed.
+
+```js
+const session = new Session({
```
paulomarg, line 50: While this isn't hard to do, I feel like it might be cleaner if we export a helper like

```ts
const session = shopify.session.customAppSession({shop: 'my-shop'});
```

that way apps don't need to create fields in the session that will just get ignored.

paulomarg (approved):
Just a couple of nits, but overall makes sense to me.

PR #724, opened 2023-02-16
Add changesets to project to manage releases

mkevinosullivan:

Adding changesets to create a consistent `CHANGELOG` format as `shopify-app-js` and to automate the publishing.

See above ☝🏻


RELEASING.md
```diff
@@ -1,77 +1,93 @@
...
-1. Commit the files to the repo
+1. Checkout the `changeset-release/next` branch
 
    ```shell
-   git commit -m "Packaging for release vX.Y.Z-rcN"
+   git checkout changeset-release/next
    ```
 
-1. Tag the commit with an _annotated_ tag (required for the `--follow-tags` of the `push` command). Note that it's very important that the tag includes `-rc` as part of the tag name - this is how `shipit` will recognize that it's a release candidate, not a general release.
+1. Update the version string in the `lib/version.ts` file to match the version in the `package.json` file.
```
paulomarg, line 61: I think we'll also need to update this file with the appropriate RC version, and changesets is a bit prescriptive about how it should be done, right?

```suggestion
1. Update the version string in the `lib/version.ts` file to match the version in the `package.json` file. Make sure to add an `-rc.X` to the version, like so:

``
7.0.0-rc.1
``

```

paulomarg (approved):
Just nits :)

.github/workflows/next-release.yml
```diff
@@ -0,0 +1,35 @@
+name: Release from next branch
```
cquemin, line 1: Can we be more explicit.
Next release and release are a bit confusing I find.
Can we agree that: 
- a `release` is from the `main` branch.
- a `release-candidate` is from the `release-candidate' branch for example? 

in both case they would be a concept of next release. That's  I think it would be less confusing that way?

RELEASING.md
cquemin: I Agree!

RELEASING.md
```diff
@@ -1,77 +1,93 @@
...
 
    ```shell
-   git push origin shopify_api_next --follow-tags
+   yarn changeset pre exit
    ```
 
-1. Login to `shipit`, search for the _next_ environment for `shopify-api-js`, and press _Deploy_ on the appropriate commit (the commit description will be the version number).
+   And commit the changed files.
+
+1. Merge the `next` branch into `main`. This _should_ update the relevant `CHANGELOG.md` files on `main` with the changes from the release candidates.
```
cquemin, line 91: Is that to be done manually ? `Changeset` does not do it automatically ?

cquemin (commented):
A request from to move away from `next` and be explicit about `release-candidate`

RELEASING.md
mkevinosullivan: `changeset` does not do what automatically?

.github/workflows/next-release.yml
mkevinosullivan: @cquemin You want to have a `release-candidate` branch instead of `next` branch for the release candidate releases?  I'm not fussed about the name, but we should be consistent across all the repos

RELEASING.md
```diff
@@ -1,15 +1,49 @@
 # Releasing shopify-api-js
 
-- Check the Semantic Versioning page for info on how to version the new release: [http://semver.org](http://semver.org)
+## General overview of the release process
+
+The `shopify-api-js` repo uses `changesets` to track and update the `CHANGELOG.md` file, as well as to publish to NPM.
+
+### For fixes and new functionality that are being published _without_ using any pre-releases
```
cquemin, line 7: Should this be `release candidates` instead of pre-releases for consistency?

RELEASING.md
```diff
@@ -1,15 +1,49 @@
...
+
+  1. The developer creates a branch with their change using `main` as the base branch, opens a PR to obtain the necessary feedback, makes any required updates and obtains approval from the maintainers.
+
+  1. The developer merges their PR, which merges their branch into `main`. This triggers the `main-release.yml` workflow.
+
+  1. The `main-release.yml` workflow will create a branch called `changeset-release/main` and a corresponding PR called `Packages for release` (or update the branch and PR if they already exist). This branch/PR contains the necessary updates to the `CHANGELOG.md` file that will be included in the release, as well as any updates (e.g., version) to the `package.json` file.
+
+  1. When ready to release, merging the `Packages for release` PR will update `main` with the contents of the `changeset-release/main` branch, and will automatically publish the package(s) to NPM.
+
+### If pre-releases are used to obtain feedback from the community prior to an upcoming release
```
cquemin, line 17: release candidates?

RELEASING.md
```diff
@@ -19,14 +53,16 @@
    git checkout changeset-release/main
    ```
 
-1. Update the version string in the `lib/version.ts` file to match the version in the `package.json` file.
+1. Update the version string in the `lib/version.ts` file to match the version in the `package.json` file in this branch.
 
-1. While the branch is checked out, edit/remove any of the comments in the changed `CHANGELOG.md` files and commit them to the `changeset-release/main` branch.
+1. While the branch is checked out, edit/remove any of the comments in the `CHANGELOG.md` files and commit them to the `changeset-release/main` branch.
```
cquemin, line 58: Is worth emphasizing that this is ONLY if needed. But not required?

RELEASING.md
```diff
@@ -19,14 +53,16 @@
...
 
-1. Once the files in the PR reflect the desired release changes, merge the `Version Packages` PR into `main` - this triggers the release.
+1. Once the files in the PR reflect the desired release changes, merge the `Packages for release` PR into `main` - this triggers the release.
 
 1. The same `changesets/action` in the `main-release.yml` workflow will call `yarn release`, which builds the packages and pushes the changed packages to `npmjs.org`.
 
+1. After the release, there will no longer be a `Packages for release` PR.  `changesets` will re-create it the next time a branch is merged into the `main` branch.
```
cquemin, line 64: a branch "with changeset changelog files presents" its not just any branch.

RELEASING.md
```diff
@@ -35,59 +71,65 @@ For significant API changes that could result in significant refactoring on the
...
+   ```
+
+1. Update the version string in the `lib/version.ts` file to match the version in the `package.json` file. Make sure to add an `-rc.X` to the version, like so:
+
+   ```text
+   7.0.0-rc.1
    ```
 
-1. Update the version string in the `lib/version.ts` file to match the version in the `package.json` file.
+1. While the branch is checked out, edit/remove any of the comments in the changed `CHANGELOG.md` files and commit them to the `changeset-release/release-candidate` branch.
```
cquemin, line 103: same comment: only if necessary.

cquemin (commented):
Few questions and comments worth looking at.

PR #730, opened 2023-02-22
Fix: custom-store-app.md copy

bangank36:

There is some typos in `custom-store-app.md` [file](https://github.com/Shopify/shopify-api-js/blob/main/docs/guides/custom-store-app.md), it reads
```
Only the `shop` parameter is required when making the API requests from a store specifc custom app 
and the value must match the stop on which the custom app is installed.
```

My propose for fixing
```
Only the `shop` parameter is required when making the API requests from a store-specific custom app 
and the value must match the shop on which the custom app is installed.
```

## Type of change

## Checklist


bangank36:
@mkevinosullivan Thanks for the heads up!

mkevinosullivan (approved):

PR #759, opened 2023-03-07
[Feature] Global configuration to set default retries

sagrawal31:

For making RestAPI or GraphQL calls, developers often have to set default retries across the application. This is redundant. 

This PR is exposing a global configuration in `ConfigParams` and is later being used in the `HttpClient`.

## Type of change

## Checklist

### DIFF
--- a/docs/reference/shopifyApi.md
+++ b/docs/reference/shopifyApi.md
@@ -18,6 +18,7 @@ const shopify = shopifyApi({
   apiVersion: ApiVersion.July22,
   isEmbeddedApp: true,
   isCustomStoreApp: false,
+  defaultRetries: 1,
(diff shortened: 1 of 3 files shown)


sagrawal31:
Signed the CLA. Not sure how to proceed.

paulomarg:
Hey, thanks for contributing, and sorry for the delay in responding.

I think this is a good idea, but there are a couple of things that we should consider before merging it:
- We should add tests to ensure that the global setting is being used by default
- We should add a test to ensure that the request-specific param overrides the global default to prevent regressions
- We should probably rename the global setting to `defaultHttpRequestTries` - we might want to refactor all of the `tries` params in the future to be `retries`, but for now they should be consistent

PR #799, opened 2023-04-04
Make REST resource pagination thread-safe

### LINKED ISSUE #740: Pagination is broken in the OO REST interface
In the documention at https://github.com/Shopify/shopify-api-js/blob/main/docs/guides/rest-resources.md#paginated-requests it is mentioned that:

> Note: these properties are not thread-safe because they're stored statically in the class. If you are using this feature to send requests in parallel, you can store the query property in a thread-safe way, since it's a plain JavaScript object.

However, it is impossible to do so because you abstract away the `Response` (that has `response.pageInfo?.nextPage`, see: https://github.com/Shopify/shopify-api-js/blob/d606b9dad669c095825f0a1e9544ce47b6f267ea/rest/base.ts#L85)

So it's not possible to do what your Docs suggest and this API client is pretty much useless for any real world application that will certainly have numerous parallel requests going on at all times.

## Example

```typescript
do {
  // note that we can't access the `response` here
  const pageProducts: Products[] = await shopify.rest.Product.all({
    ...shopify.rest.Product.NEXT_PAGE_INFO?.query,
    session,
    status: 'active',
  });

  // if we store `shopify.rest.Product.NEXT_PAGE_INFO?.query` here, it is not guaranteed to have the correct data as 
  // we might already have resolved other API requests that are running at the same time elsewhere in our event loop

} while (shopify.rest.Product.NEXT_PAGE_INFO);
```

## Solution

The current implementation with the static class variable is fundamentally wrong in an asynchronous platform like Node.js and should be changed so that you actually return the pagination related data to the caller. 

Another alternative could be that every caller creates a separate class instance and you'd use instance variables for the pagination. However, that's a bit dangerous as people might not remember to do it.

paulomarg:

Fixes #740 

As pointed out in the issue, our current support for pagination in the REST resources isn't really helpful because it's thread unsafe.

Changing the return type of the `.all()` method to be an object that contains both the data it returned before, _and_ the pagination info, so that apps can create locally scoped variables with the pagination info, making the operation thread safe.

Note: as a bonus, I'm also renaming all of the `SCREAMING_SNAKE_CASE` variables from the resources to replace them with the more idiomatic `camelCase`.

To review this, see [the commit](https://github.com/Shopify/shopify-api-js/pull/799/commits/5a68e4a567764bdec6c73e6906a5532ed29bc0be) that doesn't include the automated changes.

## Type of change

## Checklist


cquemin (approved):
This looks much better this way

PR #801, opened 2023-04-05
Change logger to be synchronous

mkevinosullivan:

The logger is now synchronous, meaning it will block the process until the log is written. This avoids the cascading effect of making functions that use the logger async also, which was confusing to users.

Makes all the necessary changes for the logger to be synchronous (avoiding `async`/`await`), including changes to the documentation.

## Type of change

## Checklist


docs/example-migration-v5-node-template-to-v6.md
paulomarg, line 1: I don't think we should update this file. Maybe we could start a v6 => v7 migration guide.

.changeset/six-crabs-give.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/shopify-api': major
+---
+
+The logger is now synchronous. This removes unnecessary `await`'s from functions that use the logger that functionally don't need to `await` anything else.
```
paulomarg, line 5: I think we should provide a little more context here, mention that:
- only apps that implement their own loggers are affected
- you can still use async logger functions, you'll just need to handle them as promises
- `addHandlers` also becomes synchronous

paulomarg (commented):
LGTM, but I think we might need a specific guide for this release.

paulomarg (approved):

PR #822, opened 2023-04-13
Add adminApiAccessToken property to config

### LINKED ISSUE #772: webhook.process validation fails in custom apps
# Issue summary

The HMAC sent in the HTTP webhook header validates against the passed `apiSecretKey`. In a custom app, the documentation indicates using the shared secret instead.
This is the relevant line.

https://github.com/Shopify/shopify-api-js/blob/2e01ac6a383db6f3c9c574571f6b175eaa0ef517/lib/webhooks/process.ts#L53

- `@shopify/shopify-api` version: 6.2.0
- Node version: 19.7.0
- Operating system: Pop!_OS 22.04 LTS x86_64

## Expected behavior
`webhook.process` should validate the HMAC correctly

## Actual behavior
webhook validation fails. 
I'm side-stepping the issue by manually setting `apiSecretKey` to the `sharedSecret` just before validation. This works but feels like an unnecessary hack.

```ts
SHOPIFY.config.apiSecretKey = CONFIG.shopify.sharedSecret;
await SHOPIFY.webhooks
	.process({
		rawBody: request.body,
		rawRequest: request,
		rawResponse: response,
	})
	.catch((error) => {
		LOGGER.error(`Error in webhook handler: ${getErrorMessage(error)}`);
	})
	.finally(() => {
		SHOPIFY.config.apiSecretKey = CONFIG.shopify.accessToken;
	});
```

### LINKED ISSUE #800: Unable to validate webhooks using private app
# Issue summary

I am made a private app for my store, however the webhooks received are unable to be validated according to debug.
If the issue is on my end, I do apologize in advance for opening up a bug report.
However I opened the report because after following the instructions provided on the repo, it seems to me that my configuration is correct, so this leads me to believe that issue might be api-side.

- `@shopify/shopify-api` version: Latest (2023-04)
- Node version: v18.15.0
- Operating system: Ubuntu 20.04 (64 Bit)

```
[shopify-api/INFO] Processing webhook request | {apiVersion: 2023-04, domain: xxxxxx.myshopify.com, topic: collections/create, webhookId: f07cc3ae-xxxx-xxxx-xxxx-43034cb9749a}
[shopify-api/DEBUG] Webhook request is well formed | {apiVersion: 2023-04, domain: xxxxxx.myshopify.com, topic: collections/create, webhookId: f07cc3ae-xxxx-xxxx-xxxx-43034cb9749a}
[shopify-api/DEBUG] Webhook validation failed | {apiVersion: 2023-04, domain: xxxxxx.myshopify.com, topic: collections/create, webhookId: f07cc3ae-xxxx-xxxx-xxxx-43034cb9749a}
'Could not validate request for topic collections/create'
```

# Expected behavior

The webhooks should be validated.

# Actual behavior

The webhook aren't being validated.

# Steps to reproduce the problem
Here's my code:
```js
export const shopify = shopifyApi({
  apiKey: process.env.SHOP_API_KEY,
  apiSecretKey: process.env.SHOP_TOKEN,     // Note: this is the API access token, NOT the API Secret Key
  apiVersion: LATEST_API_VERSION,
  isCustomStoreApp: true,                   // this MUST be set to true (default is false)
  scopes: [],
  isEmbeddedApp: false,
  hostName: hostName,
  logger: {
    level: LogSeverity.Debug,
    httpRequests: true,
  },
  restResources,
});

const session = shopify.session.customAppSession(hostName);

export async function initShopify() {
  await shopify.webhooks.addHandlers({
    COLLECTIONS_CREATE: [
      {
        deliveryMethod: DeliveryMethod.Http,
        callbackUrl: 'https://xxx.ngrok.io/webhooks/shopify',
        callback: handleWebhookRequest,
      },
    ],
  });

  try {
    await shopify.webhooks.register({ session });
  } catch (error) {
    consoleError(error);
  }
}

router.post('/shopify', express.text({ type: '*/*' }), async (req, res) => {
  try {
    console.log(req.headers)
    await shopify.webhooks.process({
      rawBody: req.body,
      rawRequest: req,
      rawResponse: res,
    });
  } catch (error) {
    consoleError(error.message);
  }
});
```

mkevinosullivan:

Currently, for custom store apps, the `apiSecretKey` is set to the Admin API access token (NOT the app's API secret key) which is used for authenticating API requests to Shopify.  However, the validation of HMAC (e.g., webhook requests) requires the app's actual API secret key, not the Admin API access token, leading to the processing of webhook events failing without a hacky workaround.

Fixes #772 
Fixes #800

Adds `adminApiAccessToken` parameter to `config` for when `isCustomStoreApp` is `true`.  This is then used for authenticating API requests.

`apiSecretKey` must now be set to the custom store app's API secret key, which is used to validate the HMAC of webhook events received from Shopify.

## Type of change

## Checklist


paulomarg:
Would it be possible to just add the new `adminApiAccessToken` property, and:

- For requests, we do
    ```ts
    accessToken = config.adminApiAccessToken || config.apiSecretKey
    ```
- For HMAC, we continue to use `apiSecretKey` to validate

lib/__tests__/config.test.ts
```diff
@@ -88,11 +88,58 @@ describe('Config object', () => {
...
+    validParams.adminApiAccessToken = 'token';
     delete (validParams as any).scopes;
 
     expect(() => validateConfig(validParams)).not.toThrow(
       ShopifyErrors.ShopifyError,
     );
+    validParams.isCustomStoreApp = false;
+  });
+
+  // ENABLE THIS TEST PRIOR TO NEXT MAJOR RELEASE
```
paulomarg, line 100: This might be a neat opportunity to add a test helper that's version-aware! So it'll skip this test prior to v8 but run it afterwards.

lib/config.ts
```diff
@@ -30,6 +31,15 @@ export function validateConfig(params: ConfigParams<any>): ConfigInterface {
   if (!('isCustomStoreApp' in params) || !params.isCustomStoreApp) {
     mandatory.push('scopes');
   }
+  // ENABLE THIS CHECK AS PART OF THE NEXT MAJOR RELEASE
+  // if ('isCustomStoreApp' in params && params.isCustomStoreApp) {
```
paulomarg, line 35: We could add an actual check here for the version, right? With this and the optional test, I'm starting to wonder if it wouldn't be nice to have a method like `deprecated` that only runs code if VERSION >= something. That would make it easy to search-replace when we're releasing a new version.

lib/webhooks/process.ts
```diff
@@ -88,7 +88,12 @@ export function process(
         }
       } else {
         log.debug('Webhook validation failed', loggingContext);
-
+        if (config.isCustomStoreApp) {
+          log.deprecated(
```
paulomarg, line 92: Nice, this will make it very easy to spot for partners!

paulomarg (approved):
Just nits, and a potential improvement that can be a follow-up.

PR #824, opened 2023-04-14
Adds check for Google Crawler attempting to crawl Auth Routes

### LINKED ISSUE #686: [6.1.0] Cannot complete OAuth process. Could not find an OAuth cookie for shop url
# Issue summary

I am trying to demonstrate how to retrieve the Shopify store access_token using the following code, and I keep receiving this error. I reviewed another issue (https://github.com/Shopify/shopify-api-js/issues/582), and it mentioned the issue should be fixed. Unfortunately, I still experience this issue.

```typescript
  shopifyAPI.auth.callback({
      isOnline: false,
      rawResponse: request,
      rawRequest: response
  })
```

## Expected behavior

to have Shopify.auth.callback({ isOnline: true, rawRequest: req, rawResponse: res, }) to not throw an error

```
  const shopifyAPI = shopifyApi({
    apiKey: apiKey,
    apiSecretKey: apiSecret,
    scopes: getShopifyAppScopes(),
    hostName: getShopifyCallbackHostname(),
    hostScheme: SHOPIFY_APP_HOST_SCHEME,
    apiVersion: SHOPIFY_APP_API_VERSION,
    isEmbeddedApp: true,
  })
  const callbackResponse = await shopifyAPI.auth.callback({
    isOnline: false,
    rawResponse: {},
    rawRequest: {}
  })

```

## Actual behavior

it throw the error : 

```
Error: Cannot complete OAuth process. Could not find an OAuth cookie for shop url: 
    at CookieNotFound.ShopifyError [as constructor] (/Volumes/work/workspaces/my-works/testing/node_modules/@shopify/shopify-api/lib/error.ts:4:3)
    at new CookieNotFound (/Volumes/work/workspaces/my-works/testing/node_modules/@shopify/shopify-api/lib/error.js:174:42)
    at Object.<anonymous> (/Volumes/work/workspaces/my-works/testing/node_modules/@shopify/shopify-api/lib/auth/oauth/oauth.ts:139:13)
    at step (/Volumes/work/workspaces/my-works/testing/node_modules/tslib/tslib.js:144:27)
    at Object.next (/Volumes/work/workspaces/my-works/testing/node_modules/tslib/tslib.js:125:57)
    at fulfilled (/Volumes/work/workspaces/my-works/testing/node_modules/tslib/tslib.js:115:62)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
```

Could someone help to take a look at this issue? Thanks.

abharvey:

~~Fixes~~ Related to #686

Users were seeing logs with multiple auth requests coming in failing on missing cookies. We determined that in Chrome google's crawler was sending additional requests to the auth and auth/callback endpoints for our test app with the user-agent "GoogleOther" which would not have cookies set nor shop and other parameters available.

Adds a check against the google crawler user agent and for missing shop params, then throws an error (based on needed discussion)

## Type of change

## Checklist

### DIFF
--- a/lib/auth/oauth/__tests__/oauth.test.ts
+++ b/lib/auth/oauth/__tests__/oauth.test.ts
@@ -552,7 +552,7 @@ describe('callback', () => {
     expect(responseCookies.shopify_app_session).toBeUndefined();
   });
 
-  test('callback returns undefined when the request is done by a bot', async () => {
+  test('callback throws an error when the request is done by a bot', async () => {
     const botRequest = {
       method: 'GET',
       url: 'https://my-test-app.myshopify.io/totally-real-request',
@@ -561,11 +561,11 @@ describe('callback', () => {
       },
     } as NormalizedRequest;
 
-    const callbackResponse = await shopify.auth.callback({
-      rawRequest: botRequest,
-    });
-
-    expect(callbackResponse).toEqual({headers: {}, session: {}});
+    await expect(
+      shopify.auth.callback({
+        rawRequest: botRequest,
+      }),
+    ).rejects.toThrow(ShopifyErrors.InvalidOAuthError);
   });
 
   test('properly updates the OAuth cookie for offline, non-embedded apps', async () => {
--- a/lib/auth/oauth/__tests__/oauth.test.ts
+++ b/lib/auth/oauth/__tests__/oauth.test.ts
@@ -141,7 +141,7 @@ describe('beginAuth', () => {
     );
   });
 
-  test('response with a 418 when the request is a bot', async () => {
+  test('response with a 410 when the request is a bot', async () => {
     request.headers['User-Agent'] = 'Googlebot';
 
     const response: NormalizedResponse = await shopify.auth.begin({
@@ -151,7 +151,7 @@ describe('beginAuth', () => {
       rawRequest: request,
     });
 
-    expect(response.statusCode).toBe(418);
+    expect(response.statusCode).toBe(410);
   });
 
   test('fails to start if the app is private', () => {
--- a/lib/auth/oauth/__tests__/oauth.test.ts
+++ b/lib/auth/oauth/__tests__/oauth.test.ts
@@ -141,6 +141,19 @@ describe('beginAuth', () => {
     );
   });
 
+  test('response with a 418 when the request is a bot', async () => {
+    request.headers['User-Agent'] = 'Googlebot';
+
+    const response: NormalizedResponse = await shopify.auth.begin({
+      shop,
+      isOnline: true,
+      callbackPath: '/some-callback',
+      rawRequest: request,
+    });
+
+    expect(response.statusCode).toBe(418);
+  });
+
   test('fails to start if the app is private', () => {
     shopify.config.isCustomStoreApp = true;
 
@@ -537,6 +550,22 @@ describe('callback', () => {
     expect(callbackResponse.session.expires?.getTime()).toBeUndefined();
 
     expect(responseCookies.shopify_app_session).toBeUndefined();
+  });
+
+  test('callback returns undefined when the request is done by a bot', async () => {
+    const botRequest = {
+      method: 'GET',
+      url: 'https://my-test-app.myshopify.io/totally-real-request',
+      headers: {
+        'User-Agent': 'Googlebot',
+      },
+    } as NormalizedRequest;
+
+    const callbackResponse = await shopify.auth.callback({
+      rawRequest: botRequest,
+    });
+
+    expect(callbackResponse).toBeUndefined();
   });
 
   test('properly updates the OAuth cookie for offline, non-embedded apps', async () => {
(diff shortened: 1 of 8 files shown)


cquemin (commented):
Seeing the same code copy/pasted twice in the same file make me think we should create a method for it. 
It would also facilitate the implementation of the final solution (throwing vs 418) since it would be implemented only once.

With that in mind, I would argue that throwing should be used for something unexpected that we can't recover from. Here it feels to me that since we can respond with 418 (? I am a teapot ???) and we know it will happen I would vote for the response code.
However did you try to see what is the behaviour from the bot? Does it "understand" 418 ?

abharvey:
> Seeing the same code copy/pasted twice in the same file make me think we should create a method for it. It would also facilitate the implementation of the final solution (throwing vs 418) since it would be implemented only once.
> 
> With that in mind, I would argue that throwing should be used for something unexpected that we can't recover from. Here it feels to me that since we can respond with 418 (? I am a teapot ???) and we know it will happen I would vote for the response code. However did you try to see what is the behaviour from the bot? Does it "understand" 418 ?

I personally love the respond with the 418 most. And I agree with avoid a throw, I normally subscribe to the "never write exceptions" camp. Ironically it was a Goggle that had convinced me.

I think whether or not the bot understands the response isn't of importance to us. The main reason for the throw was to eject from the function and stop the auth process. Ending the request with a response I think sounds better based on your comment as well as the chat with Kevin.

Regarding the duplication, I'm on the fence. Trying to be more comfortable with duplication with [AHA](https://kentcdodds.com/blog/aha-programming)

abharvey:
@cquemin  I changed my mind and extracted it. I'll have to add the response to it but now we can compare

cquemin:
> @cquemin I changed my mind and extracted it. I'll have to add the response to it but now we can compare

I think it looked a bit better :D

lib/auth/oauth/oauth.ts
```diff
@@ -39,6 +40,25 @@ export interface CallbackResponse<T = AdapterHeaders> {
...
 
+interface BotCheckArgs {
+  request: AdapterRequest;
+  shop: string;
+  log: ShopifyLogger;
+}
+
+const respondToPossibleBotRequest = (args: BotCheckArgs) => {
+  const {request, shop, log} = args;
+  if (request.headers['User-Agent']?.includes('GoogleOther') || !shop) {
```
paulomarg, line 51: I wonder if checking the shop here is the right thing to do. We have an explicit check later on which will throw a more specific error message.

Is the concern here that there might be other bots that will follow the same pattern? In that case, would it make more sense to check that we have **no** query args?

abharvey: I like this idea. I'll change that

lib/auth/oauth/__tests__/oauth.test.ts
```diff
@@ -260,7 +273,9 @@ describe('callback', () => {
     queueMockResponse(JSON.stringify(successResponse));
 
     const callbackResponse = await shopify.auth.callback({rawRequest: request});
-
+    if (!callbackResponse) {
+      fail('Callback response is undefined');
+    }
```
teddyhwang, line 278: I presume all these callbackResponse checks are unnecessary now that you've removed the possibility of an `undefined` return?

lib/auth/oauth/oauth.ts
```diff
@@ -116,9 +136,18 @@ export function callback(config: ConfigInterface) {
...
     const shop = query.get('shop')!;
 
+    const response = {} as NormalizedResponse;
+    if (isbot(request.headers['User-Agent'])) {
+      logForBot({request, log, func: 'callback'});
+      return {
+        headers: (await abstractConvertHeaders({}, adapterArgs)) as T,
+        session: {} as Session,
+      };
+    }
```
teddyhwang, line 146: can we not return the same escape as `begin` here and return a `418`?

abharvey: We cannot, we don't handle the response in the API for the callback the same way, instead we're throwing now

mkevinosullivan (approved):

lib/auth/oauth/oauth.ts
```diff
@@ -116,9 +136,17 @@ export function callback(config: ConfigInterface) {
     ).searchParams;
     const shop = query.get('shop')!;
 
+    const response = {} as NormalizedResponse;
+    if (isbot(request.headers['User-Agent'])) {
+      logForBot({request, log, func: 'callback'});
+      throw new ShopifyErrors.BotActivityDetected(
+        'Invalid OAuth callback initiated by bot',
+      );
+    }
```
abharvey, line 145: @teddyhwang New callback bot handling.
Your last comment was on an out dated diff. Lots of back and forth on this with @mkevinosullivan and @paulomarg to decide the best course of action.
This is where we've landed.

Next step is to update the express app middleware to handle this error.

PR #840, opened 2023-04-25
Add replacement behavior and trial days to usage billing

### LINKED ISSUE #770: Usage based billing ignores replacement behaviour
Hi 👋 

the [requestUsagePayment function](https://github.com/Shopify/shopify-api-js/blob/2e01ac6a383db6f3c9c574571f6b175eaa0ef517/lib/billing/request.ts#L152) ignores the `replacementBehavior` parameter of `billingConfig`. I think this is a mistake as the [appSubscriptionCreate](https://shopify.dev/docs/api/admin-graphql/2023-01/mutations/appSubscriptionCreate) makes no restrictions on the usage of this parameter. The function was created in [this commit](https://github.com/Shopify/shopify-api-js/commit/2cd1d6c08174642b70493df111d4a154b24e070c) by @lizkenyon.

Thanks,
Hagen

lizkenyon:

Fixes #770 

`replacementBehavior` and `trialDays` were documented but left out of the original mutation.

## Type of change

## Checklist

### DIFF
--- a/.changeset/curvy-gorillas-smoke.md
+++ b/.changeset/curvy-gorillas-smoke.md
@@ -1,0 +1,5 @@
+---
+'@shopify/shopify-api': patch
+---
+
+Replace `semver` with `compare-versions` lib to reduce dependency on nodejs builtin-libs
--- a/.changeset/nine-flies-drop.md
+++ b/.changeset/nine-flies-drop.md
@@ -1,0 +1,5 @@
+---
+'@shopify/shopify-api': patch
+---
+
+apiKey configuration parameter is not mandatory when isCustomStoreApp is true. Fixes 782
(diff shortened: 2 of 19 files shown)


lib/billing/request.ts
```diff
@@ -163,6 +163,8 @@ async function requestUsagePayment({
         name: plan,
         returnUrl,
         test: isTest,
+        trialDays: billingConfig.trialDays,
+        replacementBehavior: billingConfig.replacementBehavior,
```
paulomarg, line 167: Just checking - is the actual API OK with these values being present but `null`? I would imagine so, but probably worth testing out.

.changeset/weak-shoes-join.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/shopify-api': patch
+---
+
+Add trial days and replacement behavior to usage billing mutation
```
mkevinosullivan, line 5:
```suggestion
Add trial days and replacement behavior to usage billing mutation. Fixes #770
```

PR #844, opened 2023-04-27
Corrected interface for webhookhandler

nishantkyal:
Converted type of deliveryMethod to enum instead of enum value

This is to avoid the typescript error in the typescript express app

Minor change


paulomarg:
I think this is not the right thing to do - we actually want those types to be specific to the `DeliveryMethod` because the fields the API accepts for each type are different.

Thank you for contributing, but I think the type safety is actually better the way it is now because it encourages apps to use the right fields for each method.

We look forward to your contributions in the future!

PR #854, opened 2023-05-03
Add `returnUrl` to billing request function

### LINKED ISSUE #715: Support admin.shopify.com as base host
In many places (billing, authorization), URL generation using the store's domain. 
This creates unnecessary redirect, incorrect redirect URLs after the billing charge and potentially creates various problems.

### LINKED ISSUE #784: Problem getting the embedded app url programmatically
I am trying to make an ajax call to a route that begins with /api with authenticated fetch and I need the [shopify.auth.getEmbeddedAppUrl](https://github.com/Shopify/shopify-api-js/blob/main/docs/reference/auth/getEmbeddedAppUrl.md) but it fails with

`InvalidRequestError: Request does not contain a host query parameter`.

Is it because the request originated from my frontend and not from Shopify?

If I do not have access to the host parameter, how can I create a redirect URL?

Here is the relevant part of the code in `web/index.js`

```
app.get('/api/redirect', async (req, res) => {
	const redirectURL = await shopify.api.auth.getEmbeddedAppUrl({
		rawRequest: req,
		rawResponse: res,
	});
	return res.json({ redirectURL });
}
```

It would be okay (but less preferable) to redirect to the home page as [shopify.billing.request](https://github.com/Shopify/shopify-api-js/blob/main/docs/reference/billing/request.md#shopifybillingrequest) does if it is not possible to create a custom redirect URL (I was unable to obtain this URL either).

If it matters, I'm not using typescript

- `@shopify/shopify-api` version: 6.2.0
- Node version: 18.11.0
- Operating system: MacOS

DavidWittness:

Fixes #715 
Fixes #784 

When setting up a return Url to pass to Shopify, a hard coded value is used based on the `session.shop`.

This is sufficient for some, but not all use cases.

New logic is introduced so that a developer can specify an app URL to be redirected to after billing is confirmed by the merchant. This allows the developer to control the flow of the app.

It should also fix issues where a non-embedded app is getting redirected to an embedded app after billing confirmation.

If no returnUrl is passed, the unified admin URL is now used, which should eliminate an extra redirect.

## Type of change

## Checklist

### DIFF
--- a/lib/billing/request.ts
+++ b/lib/billing/request.ts
@@ -43,7 +43,7 @@ export function request(config: ConfigInterface) {
     session,
     plan,
     isTest = true,
-    returnUrl,
+    returnUrl: returnUrlParam,
   }: RequestParams): Promise<string> {
     if (!config.billing || !config.billing[plan]) {
       throw new BillingError({
@@ -61,9 +61,9 @@ export function request(config: ConfigInterface) {
 
     const appUrl = `${config.hostScheme}://${config.hostName}`;
 
-    // if provided a return URL, use it, otherwise use the embeded app URL or hosted app URL
-    const returnedUrl =
-      returnUrl || (config.isEmbeddedApp ? embeddedAppUrl : appUrl);
+    // if provided a return URL, use it, otherwise use the embedded app URL or hosted app URL
+    const returnUrl =
+      returnUrlParam || (config.isEmbeddedApp ? embeddedAppUrl : appUrl);
 
     const GraphqlClient = graphqlClientClass({config});
     const client = new GraphqlClient({session});
@@ -75,7 +75,7 @@ export function request(config: ConfigInterface) {
           billingConfig: billingConfig as BillingConfigOneTimePlan,
           plan,
           client,
-          returnUrl: returnedUrl,
+          returnUrl,
           isTest,
         });
         data = mutationOneTimeResponse.data.appPurchaseOneTimeCreate;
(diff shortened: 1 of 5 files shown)


lib/billing/request.ts
```diff
@@ -53,10 +54,17 @@ export function request(config: ConfigInterface) {
...
-    const returnUrl = buildEmbeddedAppUrl(config)(
-      hashString(`${session.shop}/admin`, HashFormat.Base64),
+    const cleanShopName = session.shop.replace('.myshopify.com', '');
+    const embeddedAppUrl = buildEmbeddedAppUrl(config)(
+      hashString(`admin.shopify.com/store/${cleanShopName}`, HashFormat.Base64),
     );
 
+    const appUrl = `${config.hostScheme}://${config.hostName}`;
+
+    // if provided a return URL, use it, otherwise use the embeded app URL or hosted app URL
```
mkevinosullivan, line 64:
```suggestion
    // if provided a return URL, use it, otherwise use the embedded app URL or hosted app URL
```

lib/billing/request.ts
```diff
@@ -53,10 +54,17 @@ export function request(config: ConfigInterface) {
...
+    const returnedUrl =
```
mkevinosullivan, line 65: Not crazy about this variable name ... `returnedUrl` tends to imply that this was returned from somewhere, where in fact it's the setting to be passed in to the graphql ... maybe `returnUrlForMutation`?

mkevinosullivan (approved):
LGTM!  Just one nit.

PR #856, opened 2023-05-04
Add deprecation notice for Node 14 support

mkevinosullivan:

Removing Node 14 support is a breaking change.

This change adds a deprecation notice indicating that Node 14 support will be removed in the next major version.  Allows us to release a patch/minor version with other fixes first.

## Type of change

## Checklist

- not applicable ~I have documented new APIs/updated the documentation for modified APIs (for public APIs)~

### DIFF
--- a/lib/__tests__/initialization.test.ts
+++ b/lib/__tests__/initialization.test.ts
@@ -48,17 +48,13 @@ describe('shopifyApi', () => {
       testRuntimeString = version;
       shopify = shopifyApi(testConfig);
       expect(abstractRuntimeString()).toStrictEqual(version);
-      if (deprecated) {
-        expect(shopify.config.logger.log).toHaveBeenLastCalledWith(
-          LogSeverity.Warning,
-          expect.stringContaining('[Deprecated | 8.0.0] Support for'),
-        );
-      } else {
-        expect(shopify.config.logger.log).not.toHaveBeenLastCalledWith(
-          LogSeverity.Warning,
-          expect.stringContaining('[Deprecated | 8.0.0] Support for'),
-        );
-      }
+      const localExpect = deprecated
+        ? expect(shopify.config.logger.log)
+        : expect(shopify.config.logger.log).not;
+      localExpect.toHaveBeenLastCalledWith(
+        LogSeverity.Warning,
+        expect.stringContaining('[Deprecated | 8.0.0] Support for'),
+      );
     });
   });
 });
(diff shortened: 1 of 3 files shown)


lib/index.ts
```diff
@@ -75,5 +77,13 @@ export function shopifyApi<T extends ShopifyRestResources>(
     )
     .catch((err) => console.log(err));
 
+  const isNode = abstractRuntimeString().startsWith('Node');
+  if (isNode && compare(process.version, '16.0.0', '<')) {
```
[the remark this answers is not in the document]
mkevinosullivan: Updated to parse the runtime string instead.

lib/index.ts
```diff
@@ -75,5 +77,17 @@ export function shopifyApi<T extends ShopifyRestResources>(
...
 
+  const nodeVersionMatches = abstractRuntimeString().match(
+    /(Node) (v\d+\.\d+\.\d+)/,
+  );
+  const isNode = nodeVersionMatches && nodeVersionMatches[1] === 'Node';
+  const nodeVersion = nodeVersionMatches ? nodeVersionMatches[2] : '';
+  if (isNode && compare(nodeVersion, '16.0.0', '<')) {
+    shopify.logger.deprecated(
+      '8.0.0',
+      `Support for ${abstractRuntimeString()} will be removed`,
```
paulomarg, line 88: Should we also say "please upgrade to version 16 or higher"?

lib/__tests__/initialization.test.ts
```diff
@@ -0,0 +1,64 @@
...
+        expect(shopify.config.logger.log).toHaveBeenLastCalledWith(
+          LogSeverity.Warning,
+          expect.stringContaining('[Deprecated | 8.0.0] Support for'),
+        );
+      } else {
+        expect(shopify.config.logger.log).not.toHaveBeenLastCalledWith(
+          LogSeverity.Warning,
+          expect.stringContaining('[Deprecated | 8.0.0] Support for'),
+        );
+      }
```
paulomarg, line 61: Totally unnecessary here, but there's a neat trick we could do to avoid the duplication:
```suggestion
const expect = deprecated ?
  expect(shopify.config.logger.log) :
  expect(shopify.config.logger.log).not;
expect.toHaveBeenLastCalledWith(
  LogSeverity.Warning,
  expect.stringContaining('[Deprecated | 8.0.0] Support for'),
);
```

paulomarg (approved):
Just a comment but not really a suggestion for change :)

PR #873, opened 2023-05-12
Support multiple handlers for same webhook topic

### LINKED ISSUE #827: Support multiple handlers for the same webhook topic
## Overview

I would like to propose some improvements to the webhook registration logic of the library. These changes aim to address the following issues:

1. **Deletion of unrelated topics during individual webhook registration:** Currently, when registering a new topic individually, the library removes existing topics with different names. For example, if a `COLLECTIONS_UPDATE` topic exists and you register an `APP_SUBSCRIPTIONS_UPDATE` topic in a separate registration call, the `COLLECTIONS_UPDATE` webhook will be deleted. This behavior is not desired, as developers may want to subscribe to multiple topics individually. It is worth noting that subscribing to multiple topics at once does not cause this issue.

2. **Limitation on registering the same topic with different URLs:** The library currently does not allow registering the same topic with different URLs. For instance, if you register an `APP_SUBSCRIPTIONS_UPDATE` topic with `domain.com/webhook`, trying to register the same topic with `domain.com/webhook-new` will result in the deletion of the previous one. This limitation has already been discussed in [issue #427](https://github.com/Shopify/shopify-api-js/issues/427), but no changes have been made. According to a response from SBD_ on [this Shopify community post](https://community.shopify.com/c/shopify-apis-and-sdks/multiple-subscription-to-webhook-using-same-app/m-p/806772#M52780), Shopify's API allows multiple subscriptions to the same webhook topic with different URLs.

To address these issues, I have implemented the necessary logic and introduced a new parameter `allowMultipleHandlers`, which is `false` by default to maintain the library's current behavior. To enable multiple handlers, you can use the following code:

```javascript
api.webhooks.register({
  session,
  allowMultipleHandlers: true,
});
```

I am happy to add tests and submit a PR if these changes align with the goals of the core contributors. You can review my code changes in my repository [here](https://github.com/mhkafadar/shopify-api-js/commit/2be7e42db49c501f1cc460bed054d9d0293141a5). Your feedback is much appreciated. Thank you.

mhkafadar:

Fixes #827 

Support multiple handlers for the same webhook topic

## Type of change

## Checklist


paulomarg:
Hey, thank you for contributing. We actually already support multiple handlers for a topic! To use them, you can just pass in an array of handlers instead of an object, and the library will ensure they're all registered / triggered when appropriate.

Please see [the webhooks docs](https://github.com/Shopify/shopify-api-js/blob/main/packages/shopify-api/docs/reference/webhooks/addHandlers.md#example) for more details.

Issue #877, opened 2023-05-16
Add context variable to webhooks process function


chr33s:
## Overview

In Cloudflare workers/pages to access environment variables you need access to the context argument passed into functions. Without this context you are limited in what you can do within the process functions in webhooks (e.g. connect to databases, access Shopify tokens etc).

### Proposal

Include a context variable to `shopify.webhooks.process(` e.g.

```js
export const onRequestPost = async (context) => {
	return shopify(context).webhooks.process({
		context: context,
		rawBody: await context.request.text(),
		rawRequest: context.request,
	});
};
```

Then pass this context into the

```js
const webhooks = {
	APP_UNINSTALLED: {
		deliveryMethod: "http",
		callbackUrl: "/",
		callback: async (
			topic,
			shop,
			body,
			webhookId,
			apiVersion,
			context,
		) => {
			const payload = JSON.parse(body);
			console.log({ apiVersion, payload, shop, topic, webhookId, context });
		}
	}
}
```

Thereby enabling workers/pages to access environment variables `context.env.DATABASE_URL` from callback.

PR #880, opened 2023-05-18
Add context argument to webhooks process function for Cloudflare workers

chr33s:

Implements: #877

Implements a context argument to webhooks process function to support cloudworker contexts and env variables.

## Type of change

## Checklist

### DIFF
--- a/.changeset/young-bobcats-impress.md
+++ b/.changeset/young-bobcats-impress.md
@@ -1,0 +1,5 @@
+---
+'@shopify/shopify-api': minor
+---
+
+context argument to webhooks process function
--- a/packages/shopify-api/docs/guides/webhooks.md
+++ b/packages/shopify-api/docs/guides/webhooks.md
@@ -19,6 +19,8 @@ const shopify = shopifyApi({
 
 // Call shopify.webhooks.addHandlers here (see examples below)
 
+## Node.js
+
 const app = express();
 
 // Register webhooks after OAuth completes
@@ -86,6 +88,7 @@ const handleWebhookRequest = async (
   webhookRequestBody: string,
   webhookId: string,
   apiVersion: string,
+  context?: any,
 ) => {
   const sessionId = shopify.session.getOfflineId(shop);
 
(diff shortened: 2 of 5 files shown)


docs/guides/webhooks.md
```diff
@@ -79,6 +82,24 @@ app.post('/webhooks', express.text({type: '*/*'}), async (req, res) => {
...
+  try {
+    await shopify.webhooks.process({
+      context: {env}, // is object or undefined
+      rawBody: await request.text(), // is a string
+      rawRequest: request,
+    });
+  } catch (error) {
+    console.log(error.message);
+  }
+}
```
paulomarg, line 100: Should we also consider the `context` argument here, and pass it into `process` as well? Maybe the type of `context` could be an object like

```suggestion
async function handleFetch(request: Request, env: unknown, context: unknown): Promise<Response> {
  try {
    await shopify.webhooks.process({
      context: {env, context}, // is object or undefined
      rawBody: await request.text(), // is a string
      rawRequest: request,
    });
  } catch (error) {
    console.log(error.message);
  }
}
```

though the name `context` here might not make a lot of sense any more? WDYT?

docs/reference/webhooks/process.md
```diff
@@ -46,6 +64,12 @@ return response;
 
 Receives an object containing:
 
+### context
+
+`unknown`
```
paulomarg, line 69: Would it be better to type this as `any` instead of `unknown` so folks aren't forced to cast it?

paulomarg (commented):
Hey, thank you for contributing, and sorry for the delayed response!

I think this is a good change, and something we should do, but I was wondering if we could make a couple of changes (or if I'm misunderstanding something!).

docs/reference/webhooks/process.md
chr33s: Unknown was used due to internal lint rule's prevents any from being used, but I'll change it to any.

paulomarg:
Seems like we're missing passing the context in the internal call to call the handlers. I think we can approve / merge once that's fixed!

chr33s:
@paulomarg fixed, good catch.

chr33s:
@paulomarg pushed to an internal fork instead of the public one, all changes on main have been pull in as well.

PR #887, opened 2023-05-25
Update check and request to return objects with id

mkevinosullivan:

Add an optional parameter to `check` and `request` to return an object with payment/subscription details, including the `id`, that can be stored and used in a subsequent `cancel` if needed.

Parameter `returnObject` is optional and defaults to `false` (current published behaviour).

See above ☝🏻 

## Type of change

## Checklist

### DIFF
--- a/lib/billing/check.ts
+++ b/lib/billing/check.ts
@@ -6,15 +6,20 @@ import {
 import {BillingError} from '../error';
 
 import {
+  AppSubscription,
   BillingCheckParams,
+  BillingCheckResponse,
+  BillingCheckResponseObject,
   CurrentAppInstallation,
   CurrentAppInstallations,
+  OneTimePurchase,
 } from './types';
 
 interface CheckInternalParams {
   plans: string[];
   client: GraphqlClient;
   isTest: boolean;
+  returnObject: boolean;
 }
 
 interface CheckInstallationParams {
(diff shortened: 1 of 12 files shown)


lib/billing/check.ts
```diff
@@ -40,11 +43,19 @@ export function check(config: ConfigInterface) {
...
-      isTest,
-    });
+    if (returnObject) {
+      return getActivePayments({
+        plans: plansArray,
+        client,
+        isTest,
+      });
+    } else {
+      return hasActivePayment({
```
paulomarg, line 53: It seems this is using the same query as `getActivePayments`. Would it simplify the code if we just check if `getActivePayments` returns anything instead of having a separate method?

lib/billing/request.ts
```diff
@@ -110,7 +113,11 @@ export function request(config: ConfigInterface) {
       });
     }
 
-    return data.confirmationUrl;
+    if (returnObject) {
+      return data as BillingRequestResponseObject;
```
paulomarg, line 117: Could we just type `data` as this in line 74?

mkevinosullivan: `BillingRequestResponseObject` doesn't have a `userErrors` property, which `RequestResponse` does...

PR #905, opened 2023-06-06
Allows setting a domain-wide cookie in the oauth flow

cmelendez:

Some OAuth flows require a domain-wide cookie instead of the usual behavior that only works within the starting domain. For example, if you start the OAuth flow in `api.example.com` and then redirect the user to `app.example.com` to finish it, you'll get a CookieNotFound error (check out issue #686).

Allows to set an optional `cookieDomain` param when calling `shopifyApi`.

## Type of change

## Checklist

- [X] I have used `yarn changeset` to create a draft changelog entry (do NOT update the `CHANGELOG.md` file manually)


cmelendez:
Forgot to add that this PR also changes the default path when setting the auth cookie.

Currently the auth cookie is valid only on exactly the *same domain* and the *same path* that started the OAuth flow, which works great when creating an app using Shopify's CLI but fails on other cases.

An example where the current config fails:

1. OAuth flow starts in `https://api.example.com/auth/start`. The SDK sets the cookie on the `api.example.com` domain and `/auth/start` path.
2. OAuth flow ends in `https://dashboard.example.com/login/oauth` with an error (CookieNotFound) because it won't be able to find the auth cookie since the domain and path doesn't match.

The PR changes the default behavior by making the auth cookie available on the root path but at the same time it allows setting a specific domain where to make the cookie readable, but if not domain is set then the current behavior of using the initial domain is kept.

arabovs:
Nice work!

jagthedrummer:
What will it take to get this merged?

paulomarg (commented):
Hey folks, thanks for waiting while we got this reviewed.

I think we can take this PR, but there are a couple of things I think we need to tweak!

Cody-traded-clothes (commented):
cool

PR #924, opened 2023-06-21
Remove privateMetafieldNamespaces Deprecated

thanhhd:
Remove privateMetafieldNamespaces Deprecated


paulomarg:
Hi, thanks for the callout! There were a few more things we'd have to do other than just removing the field, but I covered that case in #927 as well!

I'm going to close this, but please do continue to contribute 🙂

PR #927, opened 2023-06-22
Stop sending deprecated field in webhook queries

paulomarg:

The `privateMetafieldNamespaces` field was deprecated from the webhooks queries in the GraphQL API, so we need to stop sending it in order to stop triggering the API warning.

We'd already deprecated the configuration field itself, but we should also stop sending it in queries since it will just be ignored. Since I needed to check if the app was behind a certain version to do that, I added a new `versionPriorTo` method which is just `versionCompatible` inversed.

## Type of change

## Checklist

### DIFF
--- a/lib/webhooks/register.ts
+++ b/lib/webhooks/register.ts
@@ -151,16 +151,26 @@ async function getExistingHandlers(
 }
 
 function buildCheckQuery(config: ConfigInterface, endCursor: string | null) {
-  let query = queryTemplate(TEMPLATE_GET_HANDLERS, {
-    END_CURSOR: JSON.stringify(endCursor),
-  });
+  return removeDeprecatedFields(
+    config,
+    queryTemplate(TEMPLATE_GET_HANDLERS, {
+      END_CURSOR: JSON.stringify(endCursor),
+    }),
+  );
+}
+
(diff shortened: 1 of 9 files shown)


lib/webhooks/register.ts
```diff
@@ -149,10 +150,17 @@ async function getExistingHandlers(
...
 }
 
-function buildCheckQuery(endCursor: string | null) {
-  return queryTemplate(TEMPLATE_GET_HANDLERS, {
+function buildCheckQuery(config: ConfigInterface, endCursor: string | null) {
+  let query = queryTemplate(TEMPLATE_GET_HANDLERS, {
     END_CURSOR: JSON.stringify(endCursor),
   });
+
+  // The privateMetafieldNamespaces field was deprecated in the July22 version, so we need to stop sending it
```
nelsonwittwer, line 158: just a style suggestion, but if we created a private function called `removeDepricatedFields(query)` you give yourself an explanation as to why you are modifying the query as well as room to grow as more fields get invariably depreciated in the future. The a well named function like that the, your well meaning comment is moot as your intent is easy to follow by reading the code.

nelsonwittwer (approved):
Love that we are cleaning this up! Thanks!

PR #938, opened 2023-07-06
[Setting] Add july api to available versions

RobinKemna:

Shopifys newest admin api is in version 2023-07 which is not reflected in this liibrary.

Adding the new api version string to exported api versions.

## Type of change

## Checklist


paulomarg:
Thanks for contributing! There are a couple of other things that also happen when we release a new version, so it can sometimes take us a few days to add everything after the actual API version goes live to update the library.

I'm closing this PR since we've already added the new version now :)

PR #959, opened 2023-08-09
Add storefront missing headers

mllemango:
Private tokens now use the right token name for header

We are assigning the private storefront token to the wrong token name, `X-Shopify-Storefront-Access-Token`, instead of `Shopify-Storefront-Private-Token` 
Using `Shopify-Storefront-Private-Token` if the config has a private token

## Type of change

## Checklist

### DIFF
--- a/.changeset/beige-kids-build.md
+++ b/.changeset/beige-kids-build.md
@@ -1,0 +1,5 @@
+---
+'@shopify/shopify-api': patch
+---
+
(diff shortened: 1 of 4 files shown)


paulomarg (commented):
LGTM, just want to figure out if this only applies to private apps before we merge.

paulomarg (approved):

PR #962, opened 2023-08-09
Add storefront private token header

mllemango:
Reverts Shopify/shopify-api-js#961

~Adding an extra flag to specify the use of the private token~
Adding a warning line to anyone who has both a private and public token
This hopefully avoids any issues that may be caused by the new working private token headers


lib/base-types.ts
```diff
@@ -18,6 +18,7 @@ export interface ConfigParams<T extends ShopifyRestResources = any> {
   adminApiAccessToken?: string;
   userAgentPrefix?: string;
   privateAppStorefrontAccessToken?: string;
+  usePrivateToken?: boolean;
```
paulomarg, line 21: I wonder if this name is clear enough - should we call this `usePrivateStorefrontAccessToken` for extra clarity?

lib/clients/graphql/storefront_client.ts
```diff
@@ -41,13 +41,23 @@ export class StorefrontClient extends GraphqlClient {
...
     const sdkVariant = LIBRARY_NAME.toLowerCase().split(' ').join('-');
+    const usePrivateToken = this.storefrontClass().config.usePrivateToken;
+    const privateToken =
+      this.storefrontClass().config.privateAppStorefrontAccessToken;
 
+    if (!usePrivateToken && privateToken !== undefined) {
+      logger(this.storefrontClass().config).warning(
+        'Private storefront access token is ignored when usePrivateToken is false',
+      );
+    }
```
paulomarg, line 52: We should probably log this when creating the app (maybe _in addition_ to here? Not sure if that won't be annoying) to inform the developer as early as possible.

mllemango: do you mean in the templates repo? or somewhere in this lib?

paulomarg: I was thinking when the app calls `shopifyApi`, so it'll show up right as the server starts.

lib/clients/graphql/storefront_client.ts
```diff
@@ -41,13 +41,24 @@ export class StorefrontClient extends GraphqlClient {
 
   protected getApiHeaders(): HeaderParams {
     const sdkVariant = LIBRARY_NAME.toLowerCase().split(' ').join('-');
+    const privateToken =
+      this.storefrontClass().config.privateAppStorefrontAccessToken;
+
+    if (this.storefrontAccessToken && privateToken) {
+      logger(this.storefrontClass().config).warning(
+        'You have both private and public storefront access tokens. The private token will be used.',
```
mllemango, line 49: Logging here instead of earlier at `validateConfig` because the config params doesn't include the public access token at that point. I could do a blanket log.info at the earlier stage but I feel like that's a little bit spammy

paulomarg: I agree - this shouldn't be a super common case but it's nice to nudge folks to fix it!

paulomarg (approved):

PR #979, opened 2023-09-05
Allow billing request trialDays override

### LINKED ISSUE #958: Do not reset remaining days from trial on uninstall
## Overview

We have a problem, where the trial period always resets when the user reinstalls the app. I've also read that this can be a feature, however for us it's definitely a bug, since anyone can just reinstall our app to use it for free forever.

There are a few community answers to solve this, however all of them includes changing the `trialDays` when requesting the recurring application charge. As far as I can see, the current implementation takes the trial days from the billing config, which can only be specified when creating the `shopifyApp` instance.

One option would be copying the config object with the new trial days, then using `validateConfig` manually on in, then passing it to the `request` factory function to create a request function. However, this does not seem like an intended way to use this library.

Is there any other way to do this that I missed? Otherwise, I think it would be great if we could pass the `trialDays` in when using the `shopify.api.billing.request` function to implement protection against indefinite free use of an app.

paulomarg:

Closes #958 

The billing API resets trials when the app is uninstalled by design, to allow more control for apps. Therefore, it falls to the app to keep track of how many trial days to grant a merchant when they set up a subscription.

However, currently the `request` method will always apply the config's `trialDays` setting, which means apps don't get to control the number of days.

Adding a `trialDaysOverride` parameter to `request`, which will enable apps to have full control over the value that gets effectively sent to the API.

A couple of questions for the reviewer:
1. Should we just name the new param `trialDays`?
1. Should we have the `request` param extend `Partial<BillingConfig>`, so that apps can override every field when calling the method?

I went with the most explicit settings here because I was concerned it might be confusing for users whether they'd need a config at all (spoilers: they do).

## Type of change

## Checklist

### DIFF
--- a/.changeset/tasty-foxes-work.md
+++ b/.changeset/tasty-foxes-work.md
@@ -1,0 +1,5 @@
+---
+'@shopify/shopify-api': minor
+---
+
+Add new trialDaysOverride parameter to billing.request, so that apps can override the default config value.
--- a/.changeset/tasty-foxes-work.md
+++ b/.changeset/tasty-foxes-work.md
@@ -2,4 +2,4 @@
 '@shopify/shopify-api': minor
 ---
 
-Add new trialDaysOverride parameter to billing.request, so that apps can override the default config value.
+Add new trialDays parameter to billing.request, so that apps can override the default config value.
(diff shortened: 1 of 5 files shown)


nelsonwittwer (approved):
I like the flexibility of overriding this for sure! I do prefer the `trialDays` name over `trialDaysOverride` as you are overriding the default implicitly by providing the param.

Issue #875, opened 2023-05-15
Cannot retrieve Customer Addresses


dilshod:
Reopening issue #384.

# Issue summary

Retrieving customer addresses returns empty list.

- `@shopify/shopify-api` version: 2023-04
- Node version: 20.1.0
- Operating system: osx 13.3.1

# Steps to reproduce

```
const resp = await shopify.rest.CustomerAddress.all({
    session: testSession,
    customer_id: "customer id",
});
console.log(resp);
```
returns
```
{
  data: [],
  headers: {...}
}
```

## Expected behavior
```
{
  data: [{...}, {...}, ...],
  headers: {...}
}
```

Changing plural name of CustomerAddress to 'addresses' fixes it:
```
    shopify.rest.CustomerAddress.pluralName = "addresses";
```

Issue #384, opened 2022-05-13
Customer Addresses doesn't parse correctly


adameichelkraut:
# Issue summary

Write a short description of the issue here ↓

Hi! I started using the Shopify SDK and wanted to use the customer addresses API to get addresses from a customer.

## Expected behavior

I should get a list of the customer's addresses. I believe that the parsing of the `PLURAL_NAME` is not parsing the name of `addresses` that the API sends back ... but rather trying to parse `customer_addresses`.

## Actual behavior

The SDK returns an empty list, but I should expect to have a list with the customer's addresses.

Tip: include an error message (in a `<details></details>` tag) if your issue is related to an error

## Steps to reproduce the problem
```javascript
const addresses = await CustomerAddress.all({
    session: testSession,
    customer_id: "<a customer id>"
});

console.log("Addresses", addresses);
```

I noticed that if I did the below, it works though :)
```javascript
(CustomerAddress as any).PLURAL_NAME = "addresses";
const addresses = await CustomerAddress.all({
    session: testSession,
    customer_id: "<a customer id>"
});

console.log("Addresses", addresses);
```

PR #986, opened 2023-09-15
Fix CustomerAddress.all

mllemango:
Some resources such as `CustomerAddress` can have multiple names when it comes to parsing the response Shopify returns. In this case, CustomerAddress will use `customer_address` most of the time, but it can also use `addresses`. 

We should allow multiple resourceNames for this reason

Raised by https://github.com/Shopify/shopify-api-js/issues/875 and https://github.com/Shopify/shopify-api-js/issues/384

## Type of change

## Checklist


rest/base.ts
```diff
@@ -212,19 +212,22 @@ export class Base {
...
-        [],
-      );
-    }
-
-    if (data[this.resourceName]) {
-      return [this.createInstance<T>(session, data[this.resourceName])];
-    }
+    let instances: T[] = [];
+    this.resourceNames.forEach((resourceName, index) => {
+      if (data[this.pluralNames[index]] || Array.isArray(data[resourceName])) {
```
paulomarg, line 217: This scares me a little bit 😄 - should we consider changing `resourceNames` so it's something like

```ts
resourceNames = [
  {singular: 'resource', plural: 'resources'},
  {singular: 'resource2', plural: 'resources2'},
]
```

paulomarg (commented):
My comment is more of a nit, but overall this LGTM.

rest/base.ts
mllemango: yeah we can totally do that

lib/clients/rest/__tests__/resources/fake-resource-with-custom-prefix.ts
```diff
@@ -10,12 +10,9 @@ interface FakeResourceWithCustomPrefixFindArgs {
...
   public static apiVersion = LATEST_API_VERSION;
-  protected static resourceNames: string[] = [
-    'fake_resource_with_custom_prefix',
-  ];
-
-  protected static pluralNames: string[] = [
-    'fake_resource_with_custom_prefixes',
+  protected static resourceNames: {[key: string]: string}[] = [
+    {singular: 'fake_resource_with_custom_prefix'},
+    {plural: 'fake_resources_with_custom_prefixes'},
```
paulomarg, line 15: I feel like these should be in the same object?

rest/base.ts
```diff
@@ -57,8 +57,8 @@ export class Base {
   public static config: ConfigInterface;
 
   public static apiVersion: string;
-  protected static resourceNames: string[] = [];
-  protected static pluralNames: string[] = [];
+  protected static resourceNames: {[key: string]: string}[] = [];
```
paulomarg, line 60: We should probably create an interface for this so we can name the `singular` and `plural` fields.

paulomarg (approved):
Just nits!

PR #988, opened 2023-09-20
Token exchange prototype

rachel-carvalho:

This is a rough prototype for the api library to support token exchange.

Allows `shopify-api` consumers to use leverage token exchange.

## Type of change

## Checklist

### DIFF
--- a/packages/shopify-api/lib/auth/index.ts
+++ b/packages/shopify-api/lib/auth/index.ts
@@ -1,6 +1,6 @@
 import {ConfigInterface} from '../base-types';
 
-import {begin, callback} from './oauth/oauth';
+import {begin, callback, tokenExchange} from './oauth/oauth';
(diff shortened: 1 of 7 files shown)


tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "compileOnSave": false,
   "compilerOptions": {
+    "outDir": "./dist",
```
paulomarg, line 4: I'm not sure this will work because of some pre-node-16 restrictions (I think we couldn't use `"moduleResolution": "node16"` yet). Since this is very likely to be a breaking change, we should consider going ahead with this, because I hated having to build in the root folder so export paths wouldn't break.

paulomarg: Well node 14 has finally reached EOL so we could probably get away with changing that and assuming node 16 from now on. Might need to be a breaking change though.

Issue #657, opened 2022-12-29
Hard to debug downcompiled async/await syntax


patdx:
# Issue summary

The stack traces for shopify API are a bit on the obtuse side, for example I just ran into this:

- `@shopify/shopify-api` version: 6.0.2
- Node version: 18
- Operating system: Mac OS

```
Error: Cannot complete OAuth process. Could not find an OAuth cookie for shop url: example.myshopify.com
    at CookieNotFound2.ShopifyError2 [as constructor] (worker.mjs:22123:28)
    at new CookieNotFound2 (worker.mjs:22284:42)
    at Object.<anonymous> (worker.mjs:31902:25)
    at step (worker.mjs:21866:25)
    at Object.next (worker.mjs:21813:20)
    at fulfilled (worker.mjs:21784:30)
```

This is kind of compounded by the fact that some issues must be debugged in a production configuration where source maps may not be available.

We can see that the source async await is downcompiled into generator syntax:

https://unpkg.com/browse/@shopify/shopify-api@6.0.2/lib/auth/oauth/oauth.js

## Expected behavior

See if it is actually necessary to downcompile the async/await syntax in the NPM library?

If it can be preserved as is, this will give more useful stack traces.

Async/await has been supported in Node since v8.

The oldest version of Node that is still in maintenance is v14.

Additionally, is is quite easy these days for the end user to downcompile async/await syntax for a runtime that does not support it.

patdx:
Still relevant with the latest version of this library: https://unpkg.com/browse/@shopify/shopify-api@6.2.0/lib/auth/oauth/oauth.js

patdx:
This is still relevant with the latest version of this library:

https://unpkg.com/browse/@shopify/shopify-api@7.0.0/lib/auth/oauth/oauth.js

Note that Node v14 is also end of life as today, the lowest version of Node that is still actively maintained is v16, which definitely has async/await support. like [every version of Node since v8](https://node.green/#ES2017-features-async-functions). Hope you can review your build settings. Thank you :)

patdx:
Still relevant: https://unpkg.com/browse/@shopify/shopify-api@7.4.0/lib/auth/oauth/oauth.js

patdx:
I think that every platform that Shopify supports also supports async/await syntax natively. So I think the transpilation does not bring any benefits and has several downsides:

- Heavier bundle size
- Slower code
- Harder to read and debug
- Dependency on tslib

public:
[...]

PR #1007, opened 2023-10-10
Drop node 14 support and remove code deprecations

paulomarg:

Some of the dependencies we use in this repo no longer work at all on Node 14, so we must drop support for it in order to keep our dependencies up to date.

Removing support for Node 14, and dropping CI checks using it.

Because this change is inherently a major release, we're also making previous deprecations final so we can go to v8.

It's easier to review this if we ignore the removal of the auto-generated files: https://github.com/Shopify/shopify-api-js/pull/1007/files/737838dde6c06b84e250e264b46c4a983cde9d61

## Type of change

## Checklist


nelsonwittwer (approved):
So long Node 14, thanks for all the fish!

patdx:
With Node 16 as the new minimum supported version, I wonder if it may be possible to consider changing the `target` in https://github.com/Shopify/shopify-api-js/blob/main/tsconfig.base.json to "es2021", which is basically what is supported by Node 16: https://www.npmjs.com/package/@tsconfig/node16 

I think this would resolve the issue I tried to raise in https://github.com/Shopify/shopify-api-js/issues/657 a while back.

PR #1008, opened 2023-10-11
[Feature] Add GraphQL Client package

melissaluu:
As part of consolidating and updating our GraphQL clients and their features, we are introducing a new generic GraphQL client that will become the bases of all other GQL API clients. This client can also be used as a standalone client library.

Add a new `graphql-client` package to the repo. 

This library exposes a `createGraphQLClient()` factory function and the resulting `client` will include the following features:
- `fetch()`: makes a request to the provided API URL and returns the native `fetch` HTTP response
- `request()`: makes a request to the provided API URL and returns a normalized response object that includes the data returned by the API, request metadata or any errors that have occurred during the request.

## Type of change

## Checklist


packages/graphql-client/.eslintignore
```diff
@@ -0,0 +1,15 @@
...
+dist/
+dist-umd/
+schema.json
+
+graphql.schema.json
+src/types/apiTypes.ts
+**/*.d.ts
+src/clientPlus/graphql/operations.ts
+codegen.ts
+transform-graphql-jest-28-shim.js
```
paulomarg, line 15: Seems like we might be ignoring some files that don't exist here?

packages/graphql-client/.eslintrc.cjs
```diff
@@ -0,0 +1,41 @@
+module.exports = {
```
paulomarg, line 1: Ideally I'd like to use the same eslint config for all packages within this repo to make them as consistent as possible. Is that an option here?

packages/graphql-client/.prettierrc.cjs
paulomarg, line 1: Same thing as eslint here.

packages/graphql-client/README.md
```diff
@@ -0,0 +1,153 @@
...
+| ------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| config        | `ClientConfig`                                                                                                                                | Configuration for the client                                                                                                                                                                                                                                                                                                                                               |
```
paulomarg, line 30: Where is `ClientConfig` used?

melissaluu: It's not used by the code but it's meant to make initial client configs available to the client user - it's not a necessity but more of a nice to have.

packages/graphql-client/.eslintignore
melissaluu: Fixed!

packages/graphql-client/.eslintrc.cjs
melissaluu: I think so - I've removed it for now but other than just rule differences, the main thing that is different between this eslint config and the main one is that this one has `browser: true`. So far this hasn't really cause any issues but it might in the future.

packages/graphql-client/.prettierrc.cjs
melissaluu: Removed!

scottdixon (commented):
Looks great @melissaluu! 🙌

packages/graphql-client/README.md
melissaluu: It is defined in the `types.ts` but I didn't want to link directly to the file. I can also change this documentation to just show the definition directly in here or add a new section for this interface - how do we do this in the shopify-api library? Is there a preference?

packages/graphql-client/src/graphql-client.ts
```diff
@@ -0,0 +1,106 @@
...
+        ...headers,
+        ...overrideHeaders,
+      },
+      body,
+    });
+  };
+
+  const request: GraphQLClient["request"] = async (...props) => {
+    try {
+      const response = await fetch(...props);
```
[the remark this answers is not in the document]
melissaluu: The local one - I can rename it to make it clearer - any preference?

packages/graphql-client/README.md
paulomarg: I agree that linking to the code wouldn't be ideal. Could we just list out the fields here like we do for the other types?

packages/graphql-client/src/utilities.ts
```diff
@@ -0,0 +1,55 @@
+export function buildDataObjectByPath(
```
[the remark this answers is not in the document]
melissaluu: Ahh good catch! It's not used...yet 😅

packages/graphql-client/src/tests/graphql-client.test.ts
```diff
@@ -0,0 +1,500 @@
...
+              ok: true,
+              headers: new Headers({
+                "Content-Type": "application/json",
+              }),
+              json: jest.fn().mockReturnValue({ data: {}, extensions }),
+            };
+
+            fetch.mockResolvedValue(mockedSuccessResponse);
+            const response = await client.request(operation, { variables });
+            expect(response).toHaveProperty("extensions", extensions);
```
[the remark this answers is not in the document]
melissaluu: No, we don't expect an error here. In this test, we're looking to see if the normalized response returns the `extensions` data when it's available.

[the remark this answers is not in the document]
paulomarg: Yeah I was just wondering if it's worth explicitly checking it's not set.

packages/graphql-client/src/types.ts
```diff
@@ -0,0 +1,59 @@
...
+  }
+) => Promise<Response>;
+
+export interface OperationVariables {
+  [key: string]: any;
+}
+
+export interface Headers {
+  [key: string]: string;
+}
```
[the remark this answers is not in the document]
melissaluu: I didn't use it here because we don't expect devs to pass in an actual `Header` object - just a set of key value pairs for their desired headers so we can easily combine any override headers if they are passed in to the `fetch` or `request` calls. 

I can rename it to something like `HeaderValues` if that's more clear - thoughts?

packages/graphql-client/src/tests/graphql-client.test.ts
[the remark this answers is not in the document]
melissaluu: Ahh I see - I can add that.

packages/graphql-client/README.md
melissaluu: Yup - can do!

packages/graphql-client/src/utilities.ts
```diff
@@ -0,0 +1,55 @@
...
+      }
+
+      acc[key] = newObject[key];
+      return acc;
+    },
+    Array.isArray(baseObject) ? [...baseObject] : { ...baseObject }
+  );
+}
+
+export function buildCombinedDataObject(dataArray: any) {
```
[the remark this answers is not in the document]
melissaluu: Removed the code so all good now 😄

packages/graphql-client/src/types.ts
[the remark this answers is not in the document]
melissaluu: I believe I had used `HeadersInit` but I remember when I was chatting with the hydrogen team, they had issues with it. Also, taking in an actual `Headers` instance won't allow for me to do the spread when I combine init headers and override headers. 

I'll stick with this implementation for now and if we see a need or usefulness to also include `Header` instances we can update then.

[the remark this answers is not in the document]
paulomarg: Yes, agreed! I think `HeadersInit` won't be available everywhere so it's less risky to just use a plain object here.

PR #1011, opened 2023-10-12
Document Shopify type changes in v8 migration

paulomarg:
An unintentional breaking change was made to the `Shopify` type when improving it for v8 - it now requires a `ConfigParams` generic before the resources, if they're being passed explicitly to the type.

It's unlikely that apps will ever need to use this since this type is only used for the `shopifyApi` function, but we should document it to help solve breakages.


byrichardpowell (approved):
:shipit:

PR #1021, opened 2023-10-26
[Storefront API Client] Add Storefront API client package

melissaluu:
Add a new `storefront-api-client` package that enables devs to create a Storefront API specific client and can be used as a standalone client library.

Add a new `storefront-api-client` package to the repo. 

This library exposes a `createStorefrontApiClient()` factory function and the resulting `client` will include the following features:
- `fetch()`: makes a request to the provided API URL and returns the native `fetch` HTTP response
- `request()`: makes a request to the provided API URL and returns a normalized response object that includes the data returned by the API, request metadata or any errors that have occurred during the request.

## Type of change

## Checklist


packages/storefront-api-client/src/storefront-api-client.ts
```diff
@@ -0,0 +1,153 @@
...
+    }
+)): StorefrontAPIClient {
+  const currentSupportedAPIVersions = getCurrentSupportedAPIVersions();
+
+  validateRequiredStoreDomain(storeDomain);
+  validateApiVersion(currentSupportedAPIVersions, apiVersion);
+  validateRequiredAccessTokens(publicAccessToken, privateAccessToken);
+  validatePrivateAccessTokenUsage(privateAccessToken);
+
+  const trimmedStoreDomain = storeDomain.trim();
```
scottdixon, line 61: Thoughts on leaning on the URL API here? We could skip the trim, cleanup and `endsWith` check further down.

```js
const { host: domain } = new URL("https://test.myshopify.com/ ")
console.log(domain) // => test.myshopify.com
```

packages/storefront-api-client/src/storefront-api-client.ts
```diff
@@ -0,0 +1,153 @@
...
+    },
+    apiUrl: generateApiUrl(),
+    clientName,
+  };
+
+  const graphqlClient = createGraphQLClient({
+    headers: config.headers,
+    url: config.apiUrl,
+    fetchAPI: clientFetchAPI,
+  });
```
scottdixon, line 101: So clean 🙌

scottdixon (commented):
Feel like I've been in a masterclass watching this evolve. Thanks @melissaluu, looking slick!

packages/storefront-api-client/src/tests/storefront-api-client.test.ts
```diff
@@ -0,0 +1,471 @@
...
+          ).toThrow(
+            new Error(
+              "Storefront API Client: private access tokens and headers should only be used in a server-to-server implementation. Use the API public access token in nonserver environments."
+            )
+          );
+        });
+
+        it("throws an error when both public and private access tokens are provided in a server environment (window is undefined)", () => {
+          const windowSpy = jest
+            .spyOn(window, "window", "get")
```
scottdixon, line 233: Why do you spy here?

packages/storefront-api-client/src/tests/storefront-api-client.test.ts
```diff
@@ -0,0 +1,471 @@
...
+      });
+    });
+
+    describe("client config", () => {
+      it("returns a config object that includes the provided store domain", () => {
+        const client = createStorefrontAPIClient(config);
+        expect(client.config.storeDomain).toBe(config.storeDomain);
+      });
+
+      it("returns a config object that includes the provided public access token, a null private access token and a header property for the public token", () => {
```
scottdixon, line 259:
```suggestion
      it("returns a config object that includes the provided public access token and a null private access token", () => {
```

Think this might be a typo after refactoring - header checks seem further down.

packages/storefront-api-client/src/storefront-api-client.ts
melissaluu: Good call out! ✅

packages/storefront-api-client/src/tests/storefront-api-client.test.ts
melissaluu: This is to simulate a server environment - the condition of being able to provide both a public and private token can only happen in a server environment and if we didn't spy on this, the earlier check for the private token only allowed in non browser scenarios will fire and we can't never reach this error message.

packages/storefront-api-client/src/storefront-api-client.ts
```diff
@@ -0,0 +1,156 @@
+import {
+  createGraphQLClient,
+  CustomFetchAPI,
+  RequestParams as GQLClientRequestParams,
+  getCurrentSupportedAPIVersions,
```
scottdixon, line 5: Nit: Should this be `getCurrentSupportedApiVersions`?

Missed this in the `graphql-client` review, sorry!

packages/storefront-api-client/src/storefront-api-client.ts
```diff
@@ -0,0 +1,211 @@
...
+function generateGetApiUrl(
+  config: StorefrontApiClientConfig,
+  apiUrlFormatter: (version?: string) => string
+): StorefrontApiClient["getApiUrl"] {
+  return (propApiVersion?: string) => {
+    return propApiVersion ? apiUrlFormatter(propApiVersion) : config.apiUrl;
+  };
+}
+
+function generateGetGQLClientProps({
```
melissaluu, line 180: This might be universal enough to move into the base client's utils...will look more into this next week.

packages/storefront-api-client/src/storefront-api-client.ts
```diff
@@ -0,0 +1,211 @@
...
+      });
+    }
+
+    const urlApiVersion = (apiVersion ?? defaultApiVersion).trim();
+
+    return `${storeUrl}/api/${urlApiVersion}/graphql.json`;
+  };
+}
+
+function generateGetHeader(
```
melissaluu, line 161: This one also looks like it's generic enough to move into the base client utils too.

packages/storefront-api-client/package.json
```diff
@@ -0,0 +1,111 @@
...
+    "graphql",
+    "Storefront API"
+  ],
+  "files": [
+    "**/*.d.ts",
+    "**/*.d.ts.map",
+    "**/*.js",
+    "**/*.js.map",
+    "**/*.mjs",
+    "**/*.mjs.map",
```
paulomarg, line 67: Would it be easier to just include everything inside `dist`?

packages/storefront-api-client/src/storefront-api-client.ts
```diff
@@ -0,0 +1,211 @@
...
+} & (
+  | {
+      publicAccessToken?: never;
+      privateAccessToken: string;
+    }
+  | {
+      publicAccessToken: string;
+      privateAccessToken?: never;
+    }
+)): StorefrontApiClient {
```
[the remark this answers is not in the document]
melissaluu: I kept it like this because in general, by writing this out, it allows the editor code hints to show the type object instead of just the name of the interface/type. I don't feel strongly about it so I'm good to go either way - do you feel strongly about it?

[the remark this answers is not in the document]
paulomarg: I feel like this is essentially duplicating the actual type we have, so we'd need to change two places if this changed - plus, the other type doesn't have the token logic in it, which is already a small bit of discrepancy between them. I'd personally prefer for this to be the type, and we rely on autocomplete more so than type hints for calling this function.

packages/storefront-api-client/src/storefront-api-client.ts
paulomarg: Yeah that's a fair concern... If we're sure the params we want to set as part of the configs will never need to be overwritten this makes sense!

melissaluu: I'll keep it like this for now and if we get feedback from devs, we can re-evaluate 😄

paulomarg (approved):
Love it!

PR #1022, opened 2023-10-27
[WIP][Admin API Client] Add Admin API client package 

scottdixon:

Add a new admin-api-client package. Used @melissaluu's Storefront client work as the starting point.

TBD

## Type of change

## Checklist


packages/admin-api-client/src/admin-api-client.ts
```diff
@@ -0,0 +1,138 @@
...
+    clientName,
+  };
+
+  const graphqlClient = createGraphQLClient({
+    headers: config.headers,
+    url: config.apiUrl,
+    fetchAPI: clientFetchAPI,
+  });
+
+  const getHeaders: AdminAPIClient["getHeaders"] = (customHeaders) => {
```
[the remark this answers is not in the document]
melissaluu: In this case, because all the methods in this client are mostly wrappers of other functions that are either defined locally in the client or requires client info, creating factories for them seems a bit much? But like you, I don't feel too strongly about this so I'm good with either way....

packages/admin-api-client/src/admin-api-client.ts
```diff
@@ -0,0 +1,143 @@
...
+    storeDomain: trimmedStoreDomain,
+    apiVersion,
+    accessToken,
+    headers: {
+      "Content-Type": DEFAULT_CONTENT_TYPE,
+      Accept: DEFAULT_CONTENT_TYPE,
+      [ACCESS_TOKEN_HEADER]: accessToken,
+      "User-Agent": `${
+        userAgentPrefix ? `${userAgentPrefix} | ` : ""
+      }${CLIENT} v${DEFAULT_CLIENT_VERSION}`,
```
scottdixon, line 83: This is a mouthful 😅 is there a cleaner approach?

packages/admin-api-client/README.md
```diff
@@ -0,0 +1,243 @@
...
+```
+
+## Admin API Client Examples
+
+### Initialize the Admin API Client
+
+```typescript
+import {createAdminAPIClient} from '@shopify/admin-api-client';
+
+const client = createAdminAPIClient({
```
melissaluu, line 26: I think the README needs to be updated to use the newer type and function names (i.e. when we switched from `API` to `Api` for naming)

packages/admin-api-client/README.md
```diff
@@ -0,0 +1,243 @@
...
+| config        | `ClientConfig`                                                                                                                                                             | Configuration for the client                                                                                                                                                                                                                                                                                                                                               |
+| getHeaders    | `(customHeaders?: Record<string, string>) => Record<string, string>`                                                                                                       | Returns Admin API specific headers needed to interact with the API. If `customHeaders` is provided, the custom headers will be included in the returned headers object.                                                                                                                                                                                               |
```
melissaluu, line 68: nit: Our TS config discourages us from using `Record<any, any>` so in the types files, we've switched to use `{[key: string]: string}` - that doesn't necessarily need to change in the README but for the other client's, I've updated them to use the format we use in the types file.

packages/admin-api-client/package.json
```diff
@@ -0,0 +1,109 @@
...
+    "graphql",
+    "admin API"
+  ],
+  "files": [
+    "**/*.d.ts",
+    "**/*.d.ts.map",
+    "**/*.js",
+    "**/*.js.map",
+    "**/*.mjs",
+    "**/*.mjs.map",
```
melissaluu, line 65: We can probably update this to use `dist/**/*.*` since all of these files should be in the `dist` folder.

packages/admin-api-client/package.json
```diff
@@ -0,0 +1,109 @@
...
+    "!node_modules"
+  ],
+  "dependencies": {
+    "@shopify/graphql-client": "^0.6.0"
```
melissaluu, line 69: This can be updated to use `^0.7.0` - the newer version of the `graphql-client` library includes updated types and has 2 new utilities.

packages/admin-api-client/src/admin-api-client.ts
```diff
@@ -0,0 +1,143 @@
...
+    client: CLIENT,
+    currentSupportedApiVersions,
+    apiVersion,
+  });
+  validateRequiredAccessToken(accessToken);
+
+  const trimmedStoreDomain = storeDomain.trim();
+  const cleanedStoreDomain = httpRegEx.test(trimmedStoreDomain)
+    ? trimmedStoreDomain.substring(trimmedStoreDomain.indexOf("//") + 2)
+    : trimmedStoreDomain;
```
melissaluu, line 55: The `validateDomainAndGetStoreUrl()` util can help you validate and get the store URL

packages/admin-api-client/src/types.ts
```diff
@@ -0,0 +1,40 @@
+import {
+  OperationVariables,
+  Headers,
+  ClientResponse,
+  GraphQLClient,
+} from "@shopify/graphql-client";
+
+export interface AdminApiClientConfig {
```
melissaluu, line 8: We can now use the common `ApiClient` types that's available in `graphql-client` to help us simplify our types 😄

packages/admin-api-client/src/validations.ts
```diff
@@ -0,0 +1,23 @@
...
+  }
+}
+
+export function validateRequiredAccessToken(accessToken: string) {
+  if (!accessToken) {
+    throw new Error(`${CLIENT}: an access token must be provided`);
+  }
+}
+
+export function validateServerSideUsage() {
```
melissaluu, line 19: This might be useful enough to move to the common validation utils in `graphql-client`

packages/admin-api-client/rollup.config.cjs
```diff
@@ -0,0 +1,101 @@
...
+
+const config = [
+  {
+    input: mainSrcInput,
+    plugins: getPlugins({
+      minify: true,
+      tsconfig: "./tsconfig.build.umd.json",
+    }),
+    output: [
+      {
```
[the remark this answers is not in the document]
scottdixon: Good catch! I don't think so.

packages/admin-api-client/src/validations.ts
melissaluu: @scottdixon - are you planning to move this over? We can do it in a follow up PR if that's the plan but just wanted to understand if there's next steps to this comment.

packages/admin-api-client/README.md
```diff
@@ -0,0 +1,243 @@
...
+const {data, errors, extensions} = await client.request(operation, {
+  variables: {
+    id: 'gid://shopify/Product/7608002183224',
+  },
+});
+```
+
+### `createAdminApiClient()` parameters
+
+| Property            | Type             | Description                                                                                                                                                                                                                                          |
```
melissaluu, line 55: I think the README still needs to be updated regarding the new `retries` and `logger` features.

packages/admin-api-client/README.md
```diff
@@ -0,0 +1,243 @@
...
+| request       | `<TData>(operation: string, options?:`[AdminAPIClientRequestOptions](#adminapiclientrequestoptions-properties)`) => Promise<`[ClientResponse\<TData\>](#clientresponsetdata)`>` | Requests data from Admin API using the provided GQL `operation` string and [AdminAPIClientRequestOptions](#adminapiclientrequestoptions-properties) object and returns a normalized response object.                                                                                                                                                                        |
+
+## `AdminAPIClientRequestOptions` properties
+
+| Name           | Type                     | Description                                          |
+| -------------- | ------------------------ | ---------------------------------------------------- |
+| variables?     | `Record<string, any>`    | Variable values needed in the graphQL operation      |
+| apiVersion?    | `string`                 | The Admin API version to use in the API request      |
+| customHeaders? | `{[key: string]: string}` | Customized headers to be included in the API request |
```
paulomarg, line 79: I'm wondering about this name - should it be just `headers`? At this point these are the only headers the app is deliberately sending, not sure if we need to say they're custom here.

packages/admin-api-client/package.json
```diff
@@ -0,0 +1,103 @@
...
+  "name": "@shopify/admin-api-client",
+  "version": "0.0.1",
+  "description": "Shopify Admin API Client - A lightweight JS client to interact with Shopify's Admin API",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/Shopify/shopify-api-js.git"
+  },
+  "author": "Shopify",
+  "license": "MIT",
+  "main": "./dist/umd/admin-api-client.min.js",
```
paulomarg, line 11: I wonder if we should point `main` to the `cjs` index file, since we're not building UMD any more, right?

packages/admin-api-client/package.json
```diff
@@ -0,0 +1,103 @@
...
+  "module": "./dist/index.mjs",
+  "types": "./dist/admin-api-client.d.ts",
+  "exports": {
+    ".": {
+      "module": {
+        "types": "./dist/ts/index.d.ts",
+        "default": "./dist/index.mjs"
```
paulomarg, line 18: Do we want to have `require` entries for the CJS build? I think that would be the right setup here?

yarn.lock
```diff
@@ -2086,7 +2086,11 @@
 
 "@shopify/graphql-client@^0.7.0":
   version "0.7.0"
+<<<<<<< HEAD
```
paulomarg, line 2089: Oops, a conflict got left behind :D 

Just running `yarn` should fix this.

paulomarg (commented):
I think we're pretty close to done here!

packages/admin-api-client/rollup.config.cjs
```diff
@@ -0,0 +1,67 @@
...
+    commonjs(),
+    typescript({
+      tsconfig: tsconfig ? tsconfig : "./tsconfig.build.json",
+      outDir: "./dist/ts",
+    }),
+    ...(minify === true ? [terser({ keep_fnames: new RegExp("fetch") })] : []),
+  ];
+}
+
+const packageName = pkg.name.substring(1);
```
melissaluu, line 28: I think we don't need these 2 variables anymore.

packages/admin-api-client/README.md
melissaluu: We could definitely change this - it's actually defined in a common ApiClient level type so I can change it there and have all the API clients use the same interface.

melissaluu (approved):
I have a few comments but none of them are critical and can be addressed in either this PR or follow up PRs. 

Generally looking good! 👍

paulomarg:
I think other than the package.json stuff (that can be a follow up as long as it gets done), I'm ok with this going in too!

packages/admin-api-client/src/tests/admin-api-client.test.ts
```diff
@@ -0,0 +1,356 @@
...
+              ...config,
+              accessToken: undefined as any,
+            })
+          ).toThrow(
+            new Error(`Admin API Client: an access token must be provided`)
+          );
+        });
+
+        it("throws an error when run in a browser environment (window is defined)", () => {
+          global.window = {} as any;
```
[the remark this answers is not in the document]
scottdixon: Wouldn't this still be handy to catch someone removing the browser/`window` check?

packages/admin-api-client/package.json
scottdixon: I think that's happening a few lines down - let me know if I'm missing something!

packages/admin-api-client/src/tests/admin-api-client.test.ts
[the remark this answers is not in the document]
melissaluu: The test itself is fine - it's the `global.window` code I think we can remove. 

Since these tests run only in a server environment, `global.window` should already be undefined.

PR #1026, opened 2023-10-31
[GraphQL Client] Update API client domain utils

melissaluu:
Updated the common API client domain utils and standardized validation function param signatures in the `graphql-client` library.

Updated, consolidated and standardized the common API client utilities.

## Type of change

## Checklist


packages/graphql-client/src/api-client-utilities/validations.ts
```diff
@@ -1,39 +1,51 @@
...
+  storeDomain: string | undefined;
+}) {
+  try {
+    if (!storeDomain || typeof storeDomain !== "string") {
+      throw new Error();
+    }
+
+    const trimmedDomain = storeDomain.trim();
+
+    const url = trimmedDomain.startsWith("http")
```
scottdixon, line 17: I don't think we need this anymore.

melissaluu: We need it because in order for us to use a non protocol domain string (eg. `test.shopify.io` as the provided store domain), I need to check whether the provided domain has the protocol and if not, add it so that we can use the `url` API. The `url` API requires the protocol or else it throws an error.

packages/graphql-client/src/api-client-utilities/validations.ts
```diff
@@ -1,39 +1,51 @@
...
+      ? trimmedDomain
+      : `https://${trimmedDomain}`;
+
+    const { host: domain } = new URL(url);
+
+    return `https://${domain}`;
```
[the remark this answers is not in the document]
melissaluu: Where do you expect to have to filter it out? The use case I have in mind ATM is for building the API URL in the individual API clients. If there's more use cases for this utility, then I'm good to remove it!

packages/graphql-client/src/api-client-utilities/validations.ts
scottdixon: Totally missed that. Nice catch!

packages/graphql-client/src/api-client-utilities/validations.ts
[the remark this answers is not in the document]
paulomarg: Discussed it, and this works as is!

PR #1036, opened 2023-11-06
Introduce token exchange api

rezaansyed:

Fixes #0000 

## Type of change

## Checklist


packages/shopify-api/lib/auth/index.ts
```diff
@@ -1,19 +1,38 @@
...
 }
 
-export type ShopifyAuth = ReturnType<typeof shopifyAuth>;
+export type ShopifyAuth<Future extends FutureFlagOptions> = {
+  begin: ReturnType<typeof begin>;
+  callback: ReturnType<typeof callback>;
+  nonce: typeof nonce;
+  safeCompare: typeof safeCompare;
+  getEmbeddedAppUrl: ReturnType<typeof getEmbeddedAppUrl>;
+  buildEmbeddedAppUrl: ReturnType<typeof buildEmbeddedAppUrl>;
```
paulomarg, line 35: These should probably be proper types exported by the files that define the functions as well.

packages/shopify-api/lib/auth/oauth/__tests__/create-session.test.ts
```diff
@@ -0,0 +1,157 @@
...
+
+beforeEach(() => {
+  shop = 'someshop.myshopify.io';
+});
+
+describe('createSession', () => {
+  describe('when receiving an offline token', () => {
+    [{isEmbeddedApp: true}, {isEmbeddedApp: false}].forEach(
+      ({isEmbeddedApp}) => {
+        test(`creates a new offline session when embedded is ${isEmbeddedApp}`, () => {
```
paulomarg, line 21: We could use `test.each` here (untested):
```suggestion
  describe('when receiving an offline token', () => {
    test.each([true, false])(`creates a new offline session when embedded is %s`, (isEmbeddedApp) => {
```

packages/shopify-api/lib/auth/oauth/create-session.ts
```diff
@@ -0,0 +1,70 @@
...
+} from './types';
+
+export function createSession({
+  config,
+  postResponse,
+  shop,
+  state,
+}: {
+  config: ConfigInterface;
+  postResponse: RequestReturn;
```
paulomarg, line 22: Should this function take in the parsed data so it doesn't assume all POSTs are going to return the same values forever? Or is that something we can guarantee won't change from Shopify's side?

paulomarg (commented):
Love it! I'll let the team review it as well, but I can't see anything I'd change other than the `ShopifyAuth` types.

packages/shopify-api/lib/auth/oauth/create-session.ts
rezaansyed: that's a good point. I've added a function that'll convert the response to the appropriate type.

packages/shopify-api/lib/auth/oauth/__tests__/create-session.test.ts
rezaansyed: Updated the test 🙂

packages/shopify-api/lib/auth/oauth/safe-compare.ts
```diff
@@ -1,5 +1,10 @@
 import * as ShopifyErrors from '../../error';
 
+export type SafeCompare = (
+  strA: string | {[key: string]: string} | string[] | number[],
+  strB: string | {[key: string]: string} | string[] | number[],
+) => boolean;
+
 export function safeCompare(
```
paulomarg, line 8: Total nit, but we could export this as an arrow function so we don't have to repeat the types:

```suggestion
export const safeCompare: SafeCompare = (strA, strB) => {
```

paulomarg (approved):

PR #1045, opened 2023-11-08
[Storefront API Client] Update client to use common API client utils

melissaluu:
To maintain a consistent API client experience, we've moved common factories and types to the base `graphql-client` repo for easier reuse and this PR updates the `storefront-api-client` to use the common factories `generateGetGQLClientParams()` and `generateGetHeaders()` and to extend the `StorefrontApiClientConfig` interface from the  common `ApiClientConfig`.

We've also added running tests in both a `browser` and `server` environment to ensure the client will work in both environments.

- updates the client to use common factories
- updates the client types to use the common types
- enables testing in both server and browser environments
- correct the README regarding access token values

## Type of change

## Checklist


packages/storefront-api-client/src/tests/storefront-api-client.test.ts
```diff
@@ -238,38 +238,40 @@ describe("Storefront API Client", () => {
...
+                )
+              );
+            });
+          });
+        }
 
-          windowSpy.mockRestore();
-        });
+        if (typeof window === "undefined") {
+          describe("server enviornment", () => {
```
paulomarg, line 260: Typo :)
```suggestion
          describe("server environment", () => {
```

packages/storefront-api-client/src/tests/storefront-api-client.test.ts
paulomarg, line 1: I wonder if it would be cleaner to have separate files with a suite for these tests, so we don't have to hide / show tests depending on a dynamic check.

We could filter them in / out at the jest project level.

paulomarg (commented):
LGTM, but I'd consider rearranging the tests a bit.

paulomarg (approved):

PR #1051, opened 2023-11-09
[Turbo] Add build dependency to local test step to mimic CI behaviour

melissaluu:
Add a `build` dependency in the local `turbo > test` pipeline so devs can recreate the our CI steps environment.

## Type of change

## Checklist


paulomarg (approved):
Great call!

melissaluu:
> Does this mean we no longer need to do that in the workflow file now?

@paulomarg - no..this is for the local `turbo test` but we can add this to the `test:ci` pipeline and update the github workflow and remove that `build` step - let me try that!

PR #1054, opened 2023-11-10
Add GQL types for Admin API

paulomarg:

We want to type GraphQL queries automatically, so apps can easily get the return types of their queries / mutations in their TS code.

Updating the types in `@shopify/admin-api-client` so that the file created by `@shopify/api-codegen-preset` automatically causes queries to be typed, without user intervention.

This is the end result:
image

## Type of change

## Checklist


packages/admin-api-client/src/admin-api-client.ts
```diff
@@ -88,17 +89,28 @@ export function createAdminApiClient({
...
 
-  const getGQLClientParams = generateGetGQLClientParams({
+  const getGQLClientParams = generateGetGQLClientParams<AdminOperations>({
     getHeaders,
     getApiUrl,
   });
 
-  const fetch = (...props: ApiClientRequestParams) => {
-    return graphqlClient.fetch(...getGQLClientParams(...props));
+  const fetch = <Operation extends keyof AdminOperations>(
```
melissaluu, line 97: Are we able to set the `fetch` and `request` to types `ApiClientFetch` and `ApiClientRequest` respectively?

eg:
> const fetch: ApiClientFetch = <Operation extends keyof AdminOperations>(...) => {...}

melissaluu: Actually, now that I'm looking at more carefully, I think we'd need to update the `ApiClientFetch` so that we can pass in `Operations` generic type for it to work - in this case `AdminOperations` needs to be passed in too right?

paulomarg: Yeah, I believe this will do the trick:

```ts
  const fetch: ApiClientFetch<AdminOperations> = (
    ...props: Parameters<ApiClientFetch<AdminOperations>>
  ) => {
    return graphqlClient.fetch(...getGQLClientParams(...props));
  };
```

But maybe I should reintroduce the `Params` type I removed to make this clearer.

melissaluu: Yeah, that would be cleaner 🤔

packages/admin-api-client/src/admin-api-client.ts
```diff
@@ -88,17 +90,17 @@ export function createAdminApiClient({
...
   const getApiUrl = generateGetApiUrl(config, apiUrlFormatter);
 
-  const getGQLClientParams = generateGetGQLClientParams({
+  const getGQLClientParams = generateGetGQLClientParams<AdminOperations>({
     getHeaders,
     getApiUrl,
   });
 
-  const fetch = (...props: ApiClientRequestParams) => {
+  const fetch: ApiClientFetch<AdminOperations> = (...props) => {
```
melissaluu, line 98:
```suggestion
  const fetch: AdminApiClient['fetch'] = (...props) => {
```

This might make more sense to devs since the `ApiClientFetch` is a generic type for all API clients in general.

packages/storefront-api-client/src/storefront-api-client.ts
```diff
@@ -100,12 +101,12 @@ export function createStorefrontApiClient({
     getApiUrl,
   });
 
-  const fetch = (...props: ApiClientRequestParams) => {
+  const fetch: ApiClientFetch = (...props) => {
```
melissaluu, line 104: We can do the same thing here as the `AdminApiClient` 😄  - one less change for the future PR 😉

paulomarg: Already done in a separate branch for SFAPI types :)

melissaluu (approved):
Just left a README comment but other than that, all looks good!

PR #1058, opened 2023-11-15
Add Storefront API GraphQL types

paulomarg:
This PR applies the same functionality as #1054 for the Storefront API client, which will enable 

## Type of change

## Checklist


packages/storefront-api-client/README.md
```diff
@@ -338,6 +338,53 @@ if (response.ok) {
...
+1. Create a `.graphqlrc.ts` file in your root containing:
+    ```ts
+    import { ApiType, shopifyApiProject } from "@shopify/api-codegen-preset";
+
+    export default {
+      schema: "https://shopify.dev/storefront-graphql-direct-proxy",
+      documents: ["*.ts", "!node_modules"],
+      projects: {
+        default: shopifyApiProject({
+          apiType: ApiType.Storefront,
```
melissaluu, line 361: Similar to the Admin API comment but we should explicitly set the `apiVersion` here since it's a required field for the api clients anyways and not setting it will likely actually cause more problems in the long run.

PR #1066, opened 2023-11-28
Add expired checks

rezaansyed:

We will require check on whether access tokens are expired without having to worry about whether scopes match or not.

Adds an `isExpired()` and `isScopeChanged()` functions.

## Type of change

## Checklist

### DIFF
--- a/.changeset/cool-mangos-brush.md
+++ b/.changeset/cool-mangos-brush.md
@@ -1,0 +1,5 @@
+---
+"@shopify/shopify-api": major
+---
+
+Add isExpired() and isScopeChanged() functions to Session class


packages/shopify-api/lib/session/session.ts
```diff
@@ -89,18 +89,22 @@ export class Session {
...
-      (!this.expires || this.expires >= new Date())
-    ) {
-      return true;
-    }
-    return false;
+    return !scopesObject.equals(this.scope);
+  }
+
+  public isExpired(): boolean {
+    return Boolean(this.expires && this.expires < new Date());
```
paulomarg, line 107: Total nit, but I would expect this condition to already be a boolean?

paulomarg (approved):

PR #1067, opened 2023-11-29
Refactor runtime adapter abstract fetch

paulomarg:

In order to properly use the new clients in a runtime-agnostic way, we'll need to pass in a `fetch` implementation when creating them.

While we already have this logic in `shopify-api`, it's currently based on our `NormalizedRequest` and `NormalizedResponse` objects, which are essentially translating `fetch` for every runtime we support.

Dropping the middle man here and making `fetch` the API for `abstractFetch`. While this is a breaking change for apps using custom runtimes, it's very likely that any environments are simply doing what we do and translating `fetch` as well. If not, they can be updated to conform with the Web API pattern in the same way they're converted to our own format now.

This has the advantage of bringing us closer to the community standard rather than introducing our own, and I feel it's more likely to make it _easier_ to add new runtimes in the future.

## Type of change

## Checklist


.changeset/rich-ladybugs-own.md
```diff
@@ -0,0 +1,13 @@
+---
+"@shopify/shopify-api": major
+---
+
+> [!NOTE]
+> This change only affects apps that are using custom runtime adapters.
+> If you're using a default adapter from this package, you don't need to make this change.
```
byrichardpowell, line 7: ~~Worth giving a code example that shows what using a default adapter from this package means? Wouldn't suprise me if partners don't understand what this means, might be better to show them?~~

The migration guide is great, is it worth linking to this here?

packages/shopify-api/adapters/node/index.ts
```diff
@@ -8,18 +10,19 @@ import {
...
   nodeConvertRequest,
   nodeConvertIncomingResponse,
   nodeConvertAndSendResponse,
   nodeConvertAndSetHeaders,
   nodeRuntimeString,
 } from './adapter';
 
-setAbstractFetchFunc(nodeFetch);
+// For the purposes of this package, fetch correctly implements everything we need
+setAbstractFetchFunc(fetch as any as AbstractFetchFunc);
```
byrichardpowell, line 25: Is there anyway to improve the type of `AbstractFetchFunc` so that we don't need to cast as any?

byrichardpowell (approved):
A few small comments, but approving so you don't need a second review.

lizkenyon (approved):

PR #1068, opened 2023-11-29
Use new GraphQL clients in shopify-api

paulomarg:

Now that `@shopify/storefront-api-client` and `@shopify/admin-api-client` are released, we want to start using them in this package.

Underpinning the existing clients with the new ones. As part of this change, I also did a bit of refactoring on the tests and rearranged the files inside `lib/clients` to match the exported object more closely.

Since this is already a big change (mostly due to tests moving around), I'm setting up an update of the REST client in a different PR.

## Type of change

## Checklist


packages/shopify-api/docs/migrating-to-v9.md
```diff
@@ -69,3 +85,43 @@ const convertFetch: AbstractFetchFunc = (url, init) => {
...
+  },
+});
+console.log(response.body, response.headers);
+```
+
+After:
+
+```ts
+const client = shopify.clients.admin.graphql({session});
+const response = client.query(
```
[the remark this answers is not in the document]
paulomarg: No, I should fix this :)

packages/graphql-client/src/graphql-client/types.ts
```diff
@@ -26,12 +26,15 @@ export interface GQLExtensions {
...
+export interface FetchResponseBody<TData = any> {
   data?: Partial<TData>;
-  errors?: ResponseErrors;
   extensions?: GQLExtensions;
 }
 
+export interface ClientResponse<TData = any> extends FetchResponseBody<TData> {
+  errors?: ResponseErrors;
+}
+
```
paulomarg, line 37: @melissaluu nit-y question: do you think these types should be in the API utilities instead of in graphql-client?

packages/shopify-api/lib/__test-helpers__/to-match-made-http-request.ts
```diff
@@ -0,0 +1,80 @@
...
+            : matchingRequest!.body;
+        expect(requestBody).toMatchObject(data);
+      }
+    } else {
+      expect(matchingRequest!.body).toBeFalsy();
+    }
+  }
+
+  return {
+    message: () => `expected to have seen the right HTTP requests`,
```
byrichardpowell, line 77: Could this error message be improved? E.g: What is the "right" HTTP requests? 

In a test helper like this I'd expect to see `expected` and `saw`

packages/shopify-api/lib/index.ts
```diff
@@ -48,22 +44,20 @@ export interface Shopify<
...
-  Resources extends DefaultedResources<Params['restResources']>,
+  Params extends ConfigParams<Resources, Future>,
+  Resources extends ShopifyRestResources,
   Future extends FutureFlagOptions,
 >({
   future,
+  restResources,
   ...config
-}: {future?: Future} & Params): Shopify<
+}: {future?: Future; restResources?: Resources} & Params): Shopify<
```
byrichardpowell, line 54: Why is this change a future flag? I assume it's not a breaking change, is it just that we want to eventually remove some code?

packages/graphql-client/src/graphql-client/types.ts
paulomarg: Ok, thanks!

packages/shopify-api/docs/migrating-to-v9.md
```diff
@@ -8,10 +8,26 @@ In this major version, our focus has been to integrate the `@shopify/shopify-api
 > This change is only breaking for apps that have created a custom runtime adapter that calls `setAbstractFetchFunc`.
 > All our existing adapters were already updated accordingly.
 
+The new clients provide the same level of functionality as the current ones, plus some other advantages:
+- You can combine them with the `@shopify/api-codegen-preset` package to automatically add types for the variables and return objects
+- Better support for the Storefront API
+- We've refactored the `query` method API so that it's closer to common GraphQL clients (e.g. Apollo) to make them feel more familiar:
+    ```ts
+    const client = shopify.clients.admin.graphql({ ... });
```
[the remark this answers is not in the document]
paulomarg: Whoops, you're absolutely right. I forgot to fix this back when I reverted.

.changeset/fifty-buckets-love.md
```diff
@@ -0,0 +1,8 @@
+---
+"@shopify/shopify-api": minor
+"@shopify/storefront-api-client": patch
+"@shopify/admin-api-client": patch
+"@shopify/graphql-client": patch
+---
+
+Use the new GraphQL API clients in shopify-api to use all of the latest features, including automatic types for query / mutation return object and variables.
```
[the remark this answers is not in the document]
paulomarg: I think the best thing here would be to add a documentation page that goes over that step by step. That page can include a video / gif of the experience, plus instructions on setting it up. Maybe we can make that a follow up and ensure we link there from the changelog / migration guide?

packages/shopify-api/lib/index.ts
paulomarg: This is actually not a future flag, but there were some types here that were a bit messy, and I'd already had to spend some time fixing them before rolling back. I figured we could keep the fixes around - this is mostly for the benefit of improving `loadRestResources`.

packages/shopify-api/lib/__test-helpers__/to-match-made-http-request.ts
paulomarg: This is actually a non-error - `pass` is set to `true` so this message will never render. I'll change it to a noop to prevent confusion.

packages/shopify-api/docs/guides/graphql-types.md
```diff
@@ -0,0 +1,70 @@
...
+In this page we'll go over the process of integrating `@shopify/api-codegen-preset` into your app to get all its benefits.
+
+## Set up
+
+The first step is to install the packages:
+
+```bash
+yarn add --dev @shopify/api-codegen-preset
+yarn add @shopify/admin-api-client @shopify/storefront-api-client
+```
```
byrichardpowell, line 20: Minor, but worth adding npm & pnpm versions?

packages/shopify-api/docs/guides/graphql-types.md
```diff
@@ -0,0 +1,70 @@
...
+
+> [!NOTE]
+> The app will need to add the clients as direct dependencies so Codegen can overload the types with the ones parsed from your code.
+
+The codegen package includes all the dependencies you'll need to run [`@graphql-codegen/cli`](https://www.npmjs.com/package/@graphql-codegen/cli), including the `graphql-codegen` script.
+
+Then, you can add a `.graphqlrc.ts` file in your app's root.
+See the [package documentation](../../../api-codegen-preset/README.md#configuration) for more details on setting up the appropriate configuration.
```
byrichardpowell, line 28: Can we provide a sample one here? Might not be practical since this package is kinda of generic?

byrichardpowell (approved):
3 small comments, but approving so you don't need another review.

Great work.

packages/shopify-api/docs/guides/graphql-types.md
```diff
@@ -0,0 +1,70 @@
...
+> To fix that, either create a query or set the [`ignoreNoDocuments` option](https://the-guild.dev/graphql/codegen/docs/config-reference/codegen-config#configuration-options) to `true`.
+
+When the script runs, it'll create a new `admin.generated.d.ts` file - or `storefront` depending on your configuration.
+The `@shopify/shopify-api` package will automatically apply the types created in those files, so your queries will have variables and return types automatically!
+
+To make your development flow faster, you can pass in the `--watch` flag to update the query types whenever you save a file:
+
+```sh
+npm run graphql-codegen -- --watch
+```
```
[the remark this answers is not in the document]
paulomarg: This is to encourage folks to run it once without watch, because codegen doesn't support caching the schema (which kind of baffles me), so we have to do it by saving it to a file.

On runs subsequent to the first one, the schema is pulled from the file and much much quicker. So this is my little hacky way of steering folks toward the more efficient way to do it.

PR #1074, opened 2023-11-29
Use new clients in e2e adapter tests

paulomarg:
This PR updates the E2E adapter tests to run some simple checks on the new graphql clients - essentially just testing that our implementation of `fetch` is correct for each environment, and it's getting correctly passed in, in a way that allows us to make actual requests.

We could see about adding more test cases when dropping the HTTP client ones, but in a way we'd be retesting the clients in a different package, which seems unnecessary.

Hiding whitespaces makes this change much easier to review!


byrichardpowell (approved):
🥳

lizkenyon (approved):

PR #1079, opened 2023-11-30
Use new clients internally

paulomarg:

With the clients being fully integrated into this package, we need to start using them internally throughout the package.

Switching all internal usage to the new clients, and changing the GraphQL proxy to also use them.

## Type of change

## Checklist


paulomarg:
This PR no longer applies because we decided not to change the GraphQL clients so drastically.

PR #1093, opened 2023-12-11
Support unified admin url when using `sanitizeShop`

zzooeeyy:
Shopify is moving towards unified admin URLs, apps should be able to install to stores when a unified admin URL is provided.

- Added util methods to convert between `unified admin url` <-> `legacy url`
- In `sanitizeShop` helper, convert URL from unified admin to legacy first before sanitizing, return sanitized legacy URL as a result.

## Type of change

## Checklist


.changeset/young-news-fail.md
```diff
@@ -0,0 +1,5 @@
+---
+"@shopify/shopify-api": minor
+---
+
+Add helpers to convert unified admin URLs, and sanitizeShop util method can now support unified admin URLs
```
paulomarg, line 5: This will be in the changelog, so just checking: is "unified admin" a term that will be clear to everyone? Not sure if we mention it in our docs.

packages/shopify-api/lib/utils/__tests__/shop-validator.test.ts
```diff
@@ -65,11 +74,16 @@ describe('sanitizeShop', () => {
...
   });
 
-  test('returns null for invalid URLs', () => {
-    const shopify = shopifyApi(testConfig());
+  [
+    INVALID_SHOP_URL_1,
+    INVALID_SHOP_URL_2,
+    VALID_SHOPIFY_HOST_BUT_NOT_VALID_UNIFIED_ADMIN,
+  ].forEach((invalidUrl) => {
+    test('returns null for invalid URLs', () => {
```
paulomarg, line 82: Slightly nitpick-y, but you can use `test.each`
```suggestion
test.each([
  INVALID_SHOP_URL_1,
  INVALID_SHOP_URL_2,
  VALID_SHOPIFY_HOST_BUT_NOT_VALID_UNIFIED_ADMIN,
])("returns null for invalid URL %s", (invalidUrl) => {
  // Test things
});
```

packages/shopify-api/lib/utils/__tests__/shop-validator.test.ts
```diff
@@ -104,6 +118,15 @@ describe('sanitizeShop', () => {
       shopify.utils.sanitizeShop(INVALID_SHOP_WITH_CUSTOM_DOMAIN_REGEX),
     ).toBe(null);
   });
+
+  VALID_UNIFIED_ADMIN_URLS.forEach(({unifiedAdminUrl, legacyAdminUrl}) => {
```
paulomarg, line 122: Same comment here :)

packages/shopify-api/lib/utils/__tests__/unified-admin-url-helper.test.ts
paulomarg, line 1: Do you think it'd be worth adding a test case for custom domains (as in, setting the `customShopDomains` setting in `shopifyApi`) here too?

packages/shopify-api/lib/utils/shop-validator.ts
```diff
@@ -17,7 +20,17 @@ export function sanitizeShop(config: ConfigInterface) {
       `^[a-zA-Z0-9][a-zA-Z0-9-_]*\\.(${domainsRegex.join('|')})[/]*$`,
     );
 
-    const sanitizedShop = shopUrlRegex.test(shop) ? shop : null;
+    const unifiedAdminRegex = new RegExp(
+      `^admin.shopify.com/store/([a-zA-Z0-9][a-zA-Z0-9-_]*)$`,
```
paulomarg, line 24: Should this take into account `admin.${domains}`? I imagined it would need to consider at least `config.customShopDomains` to work with spin URLs?

packages/shopify-api/lib/utils/unified-admin-url-helper.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    const urlComponents = unifiedAdminUrl.split('/');
+    const shopName = urlComponents[urlComponents.length - 1];
+
+    const isSpinUrl = unifiedAdminUrl.includes('spin.dev/store/');
+    if (isSpinUrl) {
+      const spinUrlComponents = unifiedAdminUrl
+        .split('.')
+        .slice(2, 5)
+        .join('.');
```
paulomarg, line 19: I have no idea how likely these are to change, but this seems slightly brittle to me.

If the URL we get is sanitized, could we use a `new URL` instead of manually parsing the domain? I would imagine that the domain would remain the same, we'd just add `${shopName}.shopify` to the beginning?

.changeset/young-news-fail.md
zzooeeyy: How about half-way to `ShopAdminUrl` ?

paulomarg: Love it.

packages/shopify-api/lib/utils/__tests__/unified-admin-url-helper.test.ts
zzooeeyy: hmm the more I think about it, the less I think we should support `customShopDomain`:

1. Unified admin only works on `.shopify.com` urls. i.e. "admin.`myshopify`.com/store/my-store" is not the same as "admin.`shopify`.com/store/my-store".
  - "admin.`myshopify`.com/store/my-store" leads to a page on a store named `Admin`
  - "admin.`shopify`.com/store/my-store" leads to the actual admin page for my-store
  
-> Therefore there is no mapping for `admin.myshopify.com/store/my-store` to a legacy admin address since that's not a real unified admin URL.

2. If we want to use `customShopDomain` for Spin urls, conversion will just as be complicated in this method because for spin unified admin url, because the mapping is:
  - "admin.`web`.constellation.fn-ln.region.com/store/my-store" to legacy url 
  - "my-store.`shopify`.constellation.fn-ln.region.com"

-> That means even if we set the customShopDomain for spin to `shopify.constellation.fn-ln.region.com`, we still have to manually manipulate the domain to convert between `web.constellation` and `shopify.constellation` for unified and legacy URLs. In this case, it's probably even harder using `customShopDomain` as a dependency to grab the domain to manipulate...


Let me know if this makes sense, happy to hop on Tuple or something to chat about it too!!

paulomarg: No, I think that makes sense. We can always add that in later if it makes sense

paulomarg (approved):
A few nits, but nothing blocking.

packages/shopify-api/lib/utils/shop-admin-url-helper.ts
```diff
@@ -0,0 +1,74 @@
...
+      }
+    } else {
+      return null;
+    }
+  };
+}
+
+// Converts my-shop.myshopify.com to admin.shopify.com/store/my-shop
+export function legacyUrlToShopAdminUrl() {
+  return (legacyAdminUrl: string): string | null => {
```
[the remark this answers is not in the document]
zzooeeyy: I wasn't quite sure about this since I modelled it after sanitizeHost()

https://github.com/Shopify/shopify-api-js/blob/bd3198d592bc1ef9e61e5e0d581ce8397702ff2f/packages/shopify-api/lib/utils/shop-validator.ts#L42-L43

I'm not familiar with TS, so I thought we needed this so it can be used as a part of `shopify.utils.method`?

PR #1105, opened 2023-12-14
Remove duplicated packages

matteodepalo:
As pointed out [here](https://github.com/Shopify/shopify-api-js/pull/1095/files#r1425807825) several packages are duplicated in the top level package.json and inside some individual packages. This PR keeps the packages at the top level.

### DIFF
--- a/.changeset/young-news-fail.md
+++ b/.changeset/young-news-fail.md
@@ -1,0 +1,5 @@
+---
+"@shopify/shopify-api": minor
+---
+
+Add helpers to convert between shop admin URLs and legacy URLs. `sanitizeShop` utility method can now support shop admin URLs.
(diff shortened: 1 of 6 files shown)


.eslintrc.cjs
```diff
@@ -8,5 +8,6 @@ module.exports = {
   rules: {
     'no-console': 0,
     '@typescript-eslint/naming-convention': 0,
+    'import/no-extraneous-dependencies': 0
```
paulomarg, line 11: Is it safe to disable this check? I wonder if it introduces any risks due to us importing implicit dependencies.

matteodepalo: You're right, it's better to be safe here. I've changed this to keep it enabled.

paulomarg (approved):

PR #1107, opened 2023-12-15
Extract REST client into `admin-api-client`

paulomarg:

Now that we have dedicated clients for the GraphQL API, we should consider doing the same for the REST API, so that apps have an easy way to interact with it as well.

Taking the `HttpClient` + `RestClient` from `shopify-api`, combining them into a single client that has an API similar to the GraphQL ones, and exporting it out of `admin-api-client`.

## Type of change

## Checklist


packages/admin-api-client/src/graphql/client.ts
```diff
@@ -78,7 +75,7 @@ export function createAdminApiClient({
   };
 
   const graphqlClient = createGraphQLClient({
-    headers: config.headers,
+    headers: config.headers ?? {},
```
melissaluu, line 78: Same as SFAPI's change - this can likely go away if we remove the change in the graphql-client util types

scottdixon (commented):
Feels so good! Thank you for working on this.

Nit: we'll also need to tweak [this line](https://github.com/Shopify/shopify-api-js/blob/main/README.md?plain=1#L21) of the readme.

lizkenyon (approved):

PR #1108, opened 2023-12-15
Underpin new REST client in `shopify-api`

paulomarg:

Following #1107, which pulled a lot of code out of this package and into `admin-api-client`, we should remove the would-be duplicate code from the package and rely on the standalone client instead.

Removing `HttpClient` completely, and changing `RestClient` to be a wrapper around the standalone client, rather than an extension of `HttpClient`.

The public API for the client stays as it was before, so this is not a breaking change.

## Type of change

## Checklist


packages/shopify-api/adapters/__e2etests__/test_apps/test-web-api-app.ts
```diff
@@ -21,9 +22,12 @@ const RED = '\x1b[31m';
 const GREEN = '\x1b[32m';
 const RESET = '\x1b[39m';
 
-const HttpClient = httpClientClass(config, 'http');
+const RestClient = restClientClass({
```
paulomarg, line 25: I think a better refactoring here would be to use `abstractFetch` instead, but a lot of these tests are actually testing error handling from the client, which would have to be dropped. Not sure what the best approach is here.

packages/shopify-api/adapters/__e2etests__/test_suite.ts
```diff
@@ -228,17 +240,6 @@ export const testSuite = [
       }),
     },
   },
-  {
```
paulomarg, line 231: This fails with a different error in the standalone client, and is tested there.

packages/shopify-api/lib/auth/oauth/oauth.ts
```diff
@@ -178,19 +179,34 @@ export function callback(config: ConfigInterface): OAuthCallback {
...
+        body: JSON.stringify(body),
+        headers: {
+          'Content-Type': DataType.JSON,
+          Accept: DataType.JSON,
+        },
+      },
+    );
+
+    if (!postResponse.ok) {
+      throwFailedRequest(
```
paulomarg, line 197: This maintains the old behaviour of throwing one of the `Http` errors if this request fails. I don't think any apps care about this, but it's non-breaking.

PR #1109, opened 2023-12-15
Improve refactoring of old GQL client methods

paulomarg:

In the previous iteration refactoring the `query` methods in the GraphQL clients, we may not have gone far enough.

The name `query` is a bit misleading, and we still had a different input / output format between `shopify-api` and the lower level clients, which may make the upper level client's interface strange as we add more things.

Since we're doing a deprecation now anyhow, we should go ahead and make all of the improvements that make sense now.

Adding a new `request` method which basically mimics the lower level clients' `request` API, but keeps the same error handling we had before so apps only need to change their calls, and not what they do afterwards. This keeps the update nice and simple.

## Type of change

## Checklist


matteodepalo (approved):
LGTM!

PR #1150, opened 2024-01-17
Update ClientResponse.data type

melissaluu:
Currently, the `ClientResponse.data` has a `Partial<TData>` type to account for scenarios when the Graph API responds with only partial data (eg. when the used access token does not have the correct access scope and permissions for certain fields within the query). However, we've received [feedback](https://github.com/Shopify/shopify-api-js/commit/194ddcf27a89edb46ad7ee348b6e8a9e65d332ff#r137199135) from the community saying that by typing `ClientResponse.data` this way, it made it more cumbersome to use the client's return because client users will need to either consistently null check the data object or typecast the response.

We've decided to update `ClientResponse.data` to the full `TData` for easier consumption of the response data object. However we acknowledge that by do so, this does create the rare occasion where the client response is incorrectly typed when the GraphQL API server responds with partial data but we feel the ergonomic gains is worth this risk.

## Type of change

## Checklist

### DIFF
--- a/.changeset/smooth-pears-boil.md
+++ b/.changeset/smooth-pears-boil.md
@@ -1,0 +1,7 @@
+---
+"@shopify/storefront-api-client": patch
+"@shopify/admin-api-client": patch
+"@shopify/graphql-client": patch
+---
+
+Remove `Partial` around the `ClientResponse.data` type for easier consumption of the client's returned typed data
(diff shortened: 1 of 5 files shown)


paulomarg (approved):
While this introduces a risk that the field might not be typed as undefined, the `Partial` also meant we could make fields optional that weren't. In the end, neither option will be 100% accurate and missing fields would likely be spotted during development, so I agree that we can opt for ergonomics here.

PR #1163, opened 2024-01-24
Token exchange docs

ragalie:
Add guides and references for token exchange
- https://github.com/Shopify/develop-app-access/issues/124
- https://github.com/Shopify/develop-app-access/issues/214

- Added Oauth guide for token exchange - [PREVIEW](https://github.com/Shopify/shopify-api-js/blob/token-exchange-docs/packages/shopify-api/docs/guides/oauth.md)
- Update homepage for `shopify.auth` reference - [PREVIEW](https://github.com/Shopify/shopify-api-js/blob/token-exchange-docs/packages/shopify-api/docs/reference/auth/README.md)
- Add reference for `shopify.auth.tokenExchange` - [PREVIEW](https://github.com/Shopify/shopify-api-js/blob/token-exchange-docs/packages/shopify-api/docs/reference/auth/tokenExchange.md)
- Add reference link in Changelog for token exchange - [PREVIEW](https://github.com/Shopify/shopify-api-js/blob/token-exchange-docs/packages/shopify-api/CHANGELOG.md#minor-changes-1)

## Type of change

## Checklist

### DIFF
--- a/packages/shopify-api/docs/guides/oauth.md
+++ b/packages/shopify-api/docs/guides/oauth.md
@@ -35,7 +35,7 @@ authenticated Shopify API queries.
 This can replace authorization code grant flow completely if you also take advantage of [Shopify managed installation](https://shopify.dev/docs/apps/auth/installation#shopify-managed-installation).
 
 Learn more about:
-  - [How token exchange works](https://shopify.dev/docs/apps/auth/get-access-tokens/token-exchange/overview)
+  - [How token exchange works](https://shopify.dev/docs/apps/auth/get-access-tokens/token-exchange)
   - [Using Shopify managed installation](https://shopify.dev/docs/apps/auth/installation#shopify-managed-installation)
   - [Configuring access scopes through the Shopify CLI](https://shopify.dev/docs/apps/tools/cli/configuration)
 


packages/shopify-api/docs/reference/auth/tokenExchange.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Node.js
+```ts
+app.get('/auth', async (req, res) => {
+  const shop = shopify.utils.sanitizeShop(req.query.shop, true)
+  const headerSessionToken = getSessionTokenHeader(request);
+  const searchParamSessionToken = getSessionTokenFromUrlParam(request);
+  const sessionToken = (headerSessionToken || searchParamSessionToken)!;
+
+  await shopify.auth.exchangeToken({
```
rachel-carvalho, line 18: This should be `shopify.auth.tokenExchange`.

https://github.com/Shopify/shopify-app-js/blob/4d47824256e46ff8e87f26ea58919a85989a7279/packages/shopify-app-remix/src/server/authenticate/admin/strategies/token-exchange.ts#L129-L133

packages/shopify-api/docs/guides/oauth.md
```diff
@@ -1,22 +1,108 @@
...
 
-> **Note**: private apps are unable to perform OAuth, because they don't require an access token to interact with API.
+This can replace authorization code grant flow completely if you also take advantage of [Shopify managed install](ZL:TODO)
+
+To learn more about:
+  - [How Token Exchange Works](ZL:TODO)
+  - [Using Shopify Managed Install for your app's installation](ZL:TODO)
+  - [Configuring Access Scopes Through the Shopify CLI](https://shopify.dev/docs/apps/tools/cli/configuration)
+
+###### Remix App
```
rezaansyed, line 39: I think this is better suited in the shopify-app-remix README. This package is only an API package for JS devs so this info may not be as useful.

rezaansyed: Ah okay no worries! Makes sense 😄

packages/shopify-api/docs/guides/oauth.md
```diff
@@ -1,22 +1,96 @@
...
+## Supported types of OAuth
+> [!TIP]
+> If you are building an embedded app, we **strongly** recommend using [Shopify managed installation](https://shopify.dev/docs/apps/auth/installation#shopify-managed-installation)
+with [token exchange](#token-exchange) instead of the authorization code grant flow.
 
-1. Start the process by calling [shopify.auth.begin](../reference/auth/begin.md) to redirect the merchant to Shopify, to ask for permission to install the app.
-1. Return the merchant to your app once they approve the app installation, by calling [shopify.auth.callback](../reference/auth/callback.md) to set up a session with an API access token.
+1. [Token Exchange](#token-exchange)
+    - Recommended for embedded apps
+    - Avoids redirect to your app to continue the OAuth flow
```
paulomarg, line 24: Not sure if this would be clear to someone without a ton of context. Could we say this instead?

```suggestion
    - Doesn't require redirects, which makes it faster and prevents flickering when loading the app
```

packages/shopify-api/docs/guides/oauth.md
```diff
@@ -1,22 +1,96 @@
...
-Once you complete the OAuth process, you'll be able to call [shopify.session.getCurrentId](../reference/session/getCurrentId.md) to fetch your session, and create API clients.
+## Token Exchange
+OAuth process by exchanging the current user's [session token](https://shopify.dev/docs/apps/auth/session-tokens) for an
+[access token](https://shopify-dev-staging2.shopifycloud.com/docs/apps/auth/access-token-types/online.md) to make
+authenticated Shopify API queries.
 
-> **Note**: private apps are unable to perform OAuth, because they don't require an access token to interact with API.
+This can replace authorization code grant flow completely if you also take advantage of [Shopify managed installation](https://shopify.dev/docs/apps/auth/installation#shopify-managed-installation)
+
+To learn more about:
```
paulomarg, line 37: I'm sorry, this is super nit-picky 😅 
```suggestion
Learn more about:
```

packages/shopify-api/docs/guides/oauth.md
```diff
@@ -1,22 +1,96 @@
...
+  - [How token exchange works](https://shopify.dev/docs/apps/auth/get-access-tokens/token-exchange/overview)
+  - [Using Shopify managed installation](https://shopify.dev/docs/apps/auth/installation#shopify-managed-installation)
+  - [Configuring access scopes through the Shopify CLI](https://shopify.dev/docs/apps/tools/cli/configuration)
+
+###### Remix App
+See ["new embedded authorization Strategy"](https://shopify.dev/docs/api/shopify-app-remix#embedded-auth-strategy) to enable this feature.
+
+###### Non-remix App
```
paulomarg, line 45: Should we highlight somewhere that the same request using `authenticatedFetch` (or a regular `fetch` when using the CDN app bridge) should be used here?

packages/shopify-api/docs/reference/auth/tokenExchange.md
```diff
@@ -0,0 +1,77 @@
...
+### Node.js
+```ts
+app.get('/auth', async (req, res) => {
+  const shop = shopify.utils.sanitizeShop(req.query.shop, true)
+  const headerSessionToken = getSessionTokenHeader(request);
+  const searchParamSessionToken = getSessionTokenFromUrlParam(request);
+  const sessionToken = (headerSessionToken || searchParamSessionToken)!;
+
+  await shopify.auth.tokenExchange({
+    sessionToken,
```
paulomarg, line 20: I totally failed to notice this before so it's on me, but couldn't we accept the request (as a `rawRequest` value) here, and pull the header / search param out ourselves?

We'd also be able to pull the shop, so the only thing they'd need to change here is the token type.

packages/shopify-api/docs/guides/oauth.md
```diff
@@ -1,22 +1,96 @@
...
+To perform [authorization code grant flow](https://shopify.dev/docs/apps/auth/get-access-tokens/authorization-code-grant), you will need to create two endpoints in your app:
+
+1. Start the process by calling [shopify.auth.begin](../reference/auth/begin.md) to redirect the merchant to Shopify, to ask for permission to install the app.
+1. Return the merchant to your app once they approve the app installation, by calling [shopify.auth.callback](../reference/auth/callback.md) to set up a session with an API access token.
+
+#### Detecting scope changes
+
+When the OAuth process is completed, the created session has a `scope` field which holds all of the access scopes that were requested from the merchant at the time.
+
+When an app's access scopes change, it needs to request merchants to go through OAuth again to renew its permissions. The library provides an easy way for you to check whether that is the case at any point in your code:
```
paulomarg, line 93: Not part of this change, but we now have `session.isScopeChanged(shopify.config.scopes)` which is a slightly cleaner way of checking.

packages/shopify-api/docs/reference/auth/tokenExchange.md
rezaansyed: We don't necessarily want to do that because we don't know at what point the app is going to perform token exchange. They may have already parsed the request and gotten the headers. IMO, it's better to be specific in these params.

paulomarg: It should be dirt cheap to get these values from the request, so I don't think the repetition would be a big deal. Not the most important thing but it creates a different API than the rest of the package (where we basically use the request for everything else).

I don't feel super strongly about this, but I'd still lean toward taking in the request.

packages/shopify-api/docs/guides/oauth.md
```diff
@@ -1,22 +1,96 @@
...
+```
+3. Start the OAuth process by calling [shopify.auth.tokenExchange](../reference/auth/tokenExchange.md) to exchange user session token to access token.
+4. Use the exchanged session token to make authenticated API queries, see [After OAuth](#after-oauth)
+
+#### Detecting scope changes
 
-## Detecting scope changes
+##### Shopify managed installation
+If your access scopes are [configured through the Shopify CLI](https://shopify.dev/docs/apps/tools/cli/configuration), scope changes will be handled by Shopify automatically.
+Learn more about [Shopify managed installation](https://shopify.dev/docs/apps/auth/installation#shopify-managed-installation). Using token exchange will guarantee that the access token retrieved will always have the latest access scopes.
```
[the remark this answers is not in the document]
rezaansyed: Yeah that makes sense. Thanks for bringing it up! 😄

packages/shopify-api/docs/reference/auth/tokenExchange.md
paulomarg: Yeah, we can revisit this while the flag is marked as unstable so there's no rush.

packages/shopify-api/docs/guides/oauth.md
```diff
@@ -65,16 +65,18 @@ shopifyApi({
...
 ##### Shopify managed installation
 If your access scopes are [configured through the Shopify CLI](https://shopify.dev/docs/apps/tools/cli/configuration), scope changes will be handled by Shopify automatically.
-Learn more about [Shopify managed installation](https://shopify.dev/docs/apps/auth/installation#shopify-managed-installation). Using token exchange will guarantee that the access token retrieved will always have the latest access scopes.
+Learn more about [Shopify managed installation](https://shopify.dev/docs/apps/auth/installation#shopify-managed-installation).
+Using token exchange will ensure that the access token retrieved will always have the latest access scopes granted by the user.
 
 ##### Not using Shopify managed installation - not recommended
 If you don't have access scopes configured through the Shopify CLI, you can still use token exchange to exchange the current user's session token for access token.
-This is not recommended because you'll have to managed both OAuth flow.
+This is not recommended because you'll have to manage both OAuth flows.
```
paulomarg, line 73: Nit: should this be an info tag?
```suggestion

> [!WARNING]
> This is not recommended because you'll have to manage both OAuth flows.
```

(as per https://github.com/orgs/community/discussions/16925)

paulomarg (approved):
Just one nit, but LGTM.

PR #1181, opened 2024-01-30
Improve ts docs

paulomarg:

The types in `shopify-api` are sparse in TSDoc comments, and the purpose of some fields isn't necessarily always clear by just looking at the field name.

Adding a whole bunch of comments.

## Type of change

## Checklist


packages/shopify-api/lib/base-types.ts
```diff
@@ -5,32 +5,105 @@ import {AuthScopes} from './auth/scopes';
...
 export type LogFunction = (severity: LogSeverity, msg: string) => void;
 
 export interface ConfigParams<
   Resources extends ShopifyRestResources = ShopifyRestResources,
   Future extends FutureFlagOptions = FutureFlagOptions,
 > {
+  /**
+   * The API key for your app.
+   *
+   * You can find this in the Partner Dashboard.
```
lizkenyon, line 20: In the past folks have gotten confused because in the partner dashboard it is called client ID

Maybe 
```suggestion
   * Also known as Client ID in your Partner Dashboard.
```

packages/shopify-api/lib/base-types.ts
```diff
@@ -5,32 +5,105 @@ import {AuthScopes} from './auth/scopes';
...
+   */
   apiKey?: string;
+  /**
+   * The API secret key for your app.
+   *
+   * You can find this in the Partner Dashboard.
```
lizkenyon, line 26:
```suggestion
   * Also known as Client Secret in your Partner Dashboard.
```

packages/shopify-api/lib/base-types.ts
```diff
@@ -5,32 +5,105 @@ import {AuthScopes} from './auth/scopes';
...
+   */
   apiSecretKey: string;
+  /**
+   * The scopes your app needs to access the API.
+   */
   scopes?: string[] | AuthScopes;
+  /**
+   * THe host name of your app.
```
lizkenyon, line 34:
```suggestion
   * The host name of your app.
```

packages/shopify-api/lib/base-types.ts
```diff
@@ -5,32 +5,105 @@ import {AuthScopes} from './auth/scopes';
...
+  /**
+   * The API version to use.
+   */
   apiVersion: ApiVersion;
+  /**
+   * Whether the app is embedded in the Shopify admin.
+   */
   isEmbeddedApp: boolean;
+  /**
+   * Whether the app is a custom store app.
```
lizkenyon, line 50: Could you add a bit more detail here. Is this partner custom apps? or merchant custom apps or both?

These are the documented [app distribution types](https://shopify.dev/docs/apps/distribution). We should reference those.

packages/shopify-api/lib/session/session.ts
```diff
@@ -75,19 +75,46 @@ export class Session {
     return obj;
   }
 
+  /**
+   * The unique identifier for the session.
+   */
   readonly id: string;
+  /**
+   * The Shopify shop domain.
```
lizkenyon, line 83: Could we show an example here? 
example.myshopify.com

I know confusion can happen when shops have custom domains

lizkenyon (approved):
Thanks for all that tedious work!

PR #1216, opened 2024-02-14
Fixes init process for Shopify environment

sle-c:

Without a dev.yml we can't use Shopify tooling effectively.

Adds `dev.yml` for Shopify environment

## Type of change

## Checklist

### DIFF
--- a/.changeset/curly-boxes-switch.md
+++ b/.changeset/curly-boxes-switch.md
@@ -1,0 +1,5 @@
+---
+"@shopify/graphql-client": patch
+---
+
+Fix GraphQL client validations for store hosts starting with http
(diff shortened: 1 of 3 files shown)


dev.yml
```diff
@@ -0,0 +1,23 @@
...
+env:
+  NODE_ENV: "development"
+
+commands:
+  server:
+    run: yarn dev
+  test:
+    syntax:
+      argument: file
+      optional: args...
```
paulomarg, line 20: Is this right? It looks kind of template-y 😄

sle-c: yeah that's a good point, I think it'll take this out because we'll use yarn for everything, dev.yml is just here for compliance

PR #1234, opened 2024-02-22
Add additional app plan information to check and subscriptions methods

### LINKED ISSUE #1186: Support accessing Line Item IDs in Active Shopify Subscriptions
Building on Issue #993, this item is for the remaining work - ability to get/query line item IDs for active subscriptions on the Shopify NPM client (@shopify/shopify-api-js) - billing helper client. The current client functionality allows for the creation of mixed-type subscriptions but lacks the capability to retrieve the IDs of individual line items (i.e. within the subscription), which is needed to actually issue usage-based charges (i.e. so we, as an app dev, can get paid). (The potential workaround is to make a raw graphQL query.)

Currently, the shopify.client.billing.subscriptions query only returns ActiveSubscriptions, which lacks this detailed information. It provides the ID of the overall subscription but not the IDs of the nested line items within it. 

```
export interface AppSubscription {
    id: string;
    name: string;
    test: boolean;
}
export interface ActiveSubscriptions {
    activeSubscriptions: AppSubscription[];
```

Please support getting/querying LineItem Ids for active Subscriptions (mixed-use), not just the IDs for the Subscription record itself.

lizkenyon:

Fixes #1186

* Add App plan line item information when returning app subscriptions
* To create app billing usage records you need access to the line item ID
* Developers should also be able to easily access the rest of the app plan information to perform other necessary updates

## Type of change

## Checklist

### DIFF
--- a/.changeset/clean-guests-develop.md
+++ b/.changeset/clean-guests-develop.md
@@ -1,0 +1,5 @@
+---
+"@shopify/api-codegen-preset": patch
+---
+
+Updated dependency on @shopify/hydrogen-codegen
--- a/.changeset/rude-keys-add.md
+++ b/.changeset/rude-keys-add.md
@@ -1,0 +1,5 @@
+---
+"@shopify/shopify-api": patch
+---
+
+Updated dependency on `jose`
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "ts-jest": "^29.1.2",
     "ts-node": "^10.9.2",
     "tslib": "^2.5.0",
-    "turbo": "^1.12.2",
+    "turbo": "^1.12.4",
     "typescript": "^5.3.2"
   },
   "dependencies": {},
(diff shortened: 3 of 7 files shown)


packages/shopify-api/lib/billing/subscriptions.ts
```diff
@@ -9,13 +9,51 @@ import {
...
+                    amount {
+                      amount
+                      currencyCode
+                    }
+                  }
+                  ... on AppSubscriptionDiscountPercentage {
+                    percentage
+                  }
+                }
+              }
```
sle-c, line 40: the type https://github.com/Shopify/shopify-api-js/pull/1234/files#diff-e366dcff14ef93ab7766ac21e2c0bf9c179386a6728a7dc92b54597e57f4dc3cR340-R353

has `price` and `interval`, do we need to include those fields here?

sle-c (approved):
minor question but overall LGTM

paulomarg (approved):

PR #1250, opened 2024-02-28
[Feature] Add authentication for fulfillment service requests

lizkenyon:

First part of  [Authentication for FulfillmentService callback requests](https://github.com/Shopify/shopify-app-js/issues/578)

* This PR adds a new API for fulfillment service.
* This currently has only one method, and that if for validating requests sent. The [validation is the same as the validation for webhooks.](https://shopify.dev/docs/apps/fulfillment/fulfillment-service-apps/manage-fulfillments#step-2-receive-fulfillment-requests-and-cancellations:~:text=Requests%20can%20be%20validated%20with%20the%20same%20HMAC%20validation%20method%20that%27s%20used%20when%20Shopify%20sends%20webhooks%20to%20apps.)

In the future we could expand on these helper functions to include, creating a fulfillment service and accepting and rejecting fulfillment requests.

## Type of change

## Checklist


packages/shopify-api/lib/fulfillment-service/types.ts
```diff
@@ -0,0 +1,32 @@
+import {AdapterArgs} from '../../runtime/types';
+
+export interface FulfillmentServiceParams extends AdapterArgs {
+  /**
+   * The raw body of the request.
+   */
+  rawBody: string;
+}
+
+export enum FulfillmentServiceErrorReason {
```
paulomarg, line 10: I think this type ended up being very similar to the flow one, right? Should we consider creating a shared utils for validating the HMAC + body combo with its own types, and then having these extend from those?

I think this could be used for:
- flow
- fulfillment services
- webhooks

more to come?

packages/shopify-api/lib/fulfillment-service/validate.ts
```diff
@@ -0,0 +1,64 @@
...
+
+import {
+  FulfillmentServiceErrorReason,
+  FulfillmentServiceInvalid,
+  FulfillmentServiceParams,
+  FulfillmentServiceValid,
+} from './types';
+
+export function validateFactory(config: ConfigInterface) {
+  return async function validate({
```
paulomarg, line 16: As per above, this could be a shared bit of code that takes in a body + request combo and does the validation.

paulomarg (commented):
LGTM, but I worry we have too much repetition for all the authentication methods that check the body HMAC.

packages/shopify-api/lib/fulfillment-service/types.ts
```diff
@@ -0,0 +1,30 @@
...
+   * The raw body of the request.
+   */
+  rawBody: string;
+}
+
+export const FulfillmentServiceValidationErrorReason = {
+  ...ValidationErrorReason,
+} as const;
+
+export type FulfillmentServiceValidationErrorReasonType =
```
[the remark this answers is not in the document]
lizkenyon: We could, this is just the same as how I had Flow. So I could change both Flow and FulfillmentService to all the base types.

PR #1252, opened 2024-02-29
Log disabled future flags at startup

paulomarg:

With the introduction of future flags, apps can opt into new features before they're officially released as default when there are breaking changes. However, right now apps would need to check the types / changelog / documentation to know when a flag is introduced.

Logging an INFO line when the library is started if a flag isn't enabled. Logs will look like this:

```
[shopify-api/INFO] Future flag v10_lineItemBilling is disabled.

  Enable this flag to use the new billing API, that supports multiple line items per plan.

```

## Type of change

## Checklist

### DIFF
--- a/packages/shopify-api/future/flags.ts
+++ b/packages/shopify-api/future/flags.ts
@@ -29,19 +29,17 @@ export function logDisabledFutureFlags(
   config: ConfigInterface,
   logger: ShopifyLogger,
 ) {
-  if (!config.future) {
+  if (!config._logDisabledFutureFlags) {
     return;
   }
 
   const logFlag = (flag: string, message: string) =>
-    `Future flag ${flag} is disabled.\n\n  ${message}\n`;
+    logger.info(`Future flag ${flag} is disabled.\n\n  ${message}\n`);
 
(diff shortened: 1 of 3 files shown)


lizkenyon (approved):

packages/shopify-api/future/flags.ts
```diff
@@ -21,3 +24,24 @@ export type FeatureEnabled<
     ? true
     : false
   : false;
+
+export function logDisabledFutureFlags(
```
lizkenyon, line 28: Related to your PR in https://github.com/Shopify/shopify-app-js/pull/679

I think it would be confusing to see these API logs when you are using shopify-app-remix. Because when you are using that you can't turn on these API flags themselves. So we would be logging these error without folks being able to action them.

lizkenyon: I am wondering if we could add some sort of flag, that we can turn off from the remix library.

paulomarg: Added a new commit that should address this problem. Also noticed a small issue with the function so improved it slightly.

lizkenyon (approved):

PR #1255, opened 2024-03-01
Refactor the validation of HMAC to be more reusable

lizkenyon:

* Flow, webhooks, and fulfillment services all validate the same way with hmac
* This pulls the common webhook validation into a reusable function

* ~Previously webhooks on returned invalid types `missing_body`, `missing_headers` and `invalid_hmac` now it will also return `missing_hmac` if the HMAC header specifically is missing. More closely aligning it to how flow works.~ We will keep the same functionality and change this at the next major version.

## Type of change

## Checklist

### DIFF
--- a/packages/shopify-api/lib/utils/types.ts
+++ b/packages/shopify-api/lib/utils/types.ts
@@ -1,0 +1,36 @@
+export enum HmacValidationType {
+  Flow = 'flow',
+  Webhook = 'webhook',
+}
+
+export enum ValidationType {
+  Flow = 'flow',
+  Webhook = 'webhook',
+}
+
+export const ValidationErrorReason = {
+  MissingBody: 'missing_body',
+  InvalidHmac: 'invalid_hmac',
+  MissingHmac: 'missing_hmac',
+} as const;
+
+export type ValidationErrorReasonType =
+  (typeof ValidationErrorReason)[keyof typeof ValidationErrorReason];
+
+export interface ValidationInvalid {
+  /**
+   * Whether the request is a valid Flow request from Shopify.
+   */
+  valid: false;
+  /**
+   * The reason why the request is not valid.
+   */
+  reason: ValidationErrorReasonType;
+}
+
+export interface ValidationValid {
+  /**
+   * Whether the request is a valid request from Shopify.
+   */
+  valid: true;
+}
--- a/packages/shopify-api/lib/utils/types.ts
+++ b/packages/shopify-api/lib/utils/types.ts
@@ -3,11 +3,6 @@ export enum HmacValidationType {
   Webhook = 'webhook',
 }
 
-export enum ValidationType {
-  Flow = 'flow',
-  Webhook = 'webhook',
-}
-
 export const ValidationErrorReason = {
   MissingBody: 'missing_body',
   InvalidHmac: 'invalid_hmac',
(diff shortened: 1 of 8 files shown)


packages/shopify-api/lib/webhooks/validate.ts
```diff
@@ -39,43 +39,24 @@ export function validateFactory(config: ConfigInterface) {
...
     }
 
-    const {hmac, valid: _valid, ...loggingContext} = webhookCheck;
-    await log.debug('Webhook request is well formed', loggingContext);
-
-    if (await validateHmacString(config, rawBody, hmac, HashFormat.Base64)) {
-      await log.debug('Webhook request is valid', loggingContext);
-      return webhookCheck;
-    } else {
-      await log.debug(
```
lizkenyon, line 57: The one thing about this refactor is we loose this specific logging about manually triggering webhooks. 
Maybe we should make some sort of override of the error messages for the general function?

packages/shopify-api/lib/validate-hmac-request.ts
```diff
@@ -0,0 +1,68 @@
...
+import {
+  ValidationErrorReason,
+  type ValidateParams,
+  ValidationErrorReasonType,
+  ValidationInvalid,
+  ValidationValid,
+  ValidationType,
+} from './types';
+
+export async function validateHmacRequest({
```
lizkenyon, line 17: Accepting suggestions for names for this.

packages/shopify-api/lib/types.ts
```diff
@@ -61,3 +61,35 @@ export enum BillingReplacementBehavior {
   ApplyOnNextBillingCycle = 'APPLY_ON_NEXT_BILLING_CYCLE',
   Standard = 'STANDARD',
 }
+
+export enum ValidationType {
```
lizkenyon, line 65: Should these types live somewhere else?

packages/shopify-api/lib/types.ts
```diff
@@ -61,3 +61,35 @@ export enum BillingReplacementBehavior {
...
+  Flow = 'flow',
+  Webhook = 'webhook',
+}
+
+export const ValidationErrorReason = {
+  MissingBody: 'missing_body',
+  InvalidHmac: 'invalid_hmac',
+  MissingHmac: 'missing_hmac',
+} as const;
```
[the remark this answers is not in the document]
lizkenyon: Webhooks and flow currently return slightly different types.

Webhooks returns `missing_headers` and flow returns `missing_hmac`.

With the refactor now webhooks would also return `missing_hmac` if the hmac header specifically is missing, and `missing_headers` if one of the other required headers is missing. (I think this makes sense to pull all of the HMAC specific validations into this new function.)

From my understanding you can't extend enums in js, so that is why I am using a const so I can extend it for webhooks to add `missing_headers`. Let me know if you have another suggestion though!

packages/shopify-api/lib/webhooks/validate.ts
lizkenyon: I added back in this specific logging for the webhook validation.

packages/shopify-api/lib/utils/types.ts
```diff
@@ -0,0 +1,36 @@
+export enum HmacValidationType {
+  Flow = 'flow',
+  Webhook = 'webhook',
+}
+
+export enum ValidationType {
+  Flow = 'flow',
+  Webhook = 'webhook',
+}
```
paulomarg, line 9: Do we need the two enums here?

paulomarg (approved):
Just some nits, but LGTM

PR #1262, opened 2024-03-05
Allow paginating customer / gift card searches

### LINKED ISSUE #1192: Support pagination when searching customers with admin rest api
## Overview/summary

When using the search method for customers in the Shopify Admin API (as provided by the @shopify/shopify-api library), the method returns only response.body, which includes the customer data. However, it does not provide access to the page_info [(Link)](https://github.com/Shopify/shopify-api-js/blob/main/packages/shopify-api/rest/admin/2023-10/customer.ts#L209) or any pagination-related information in the response, making it challenging to implement pagination for customer search results.

paulomarg:

Closes #1192 

Currently, the [`Customer`](https://shopify.dev/docs/api/admin-rest/2024-01/resources/customer#get-customers-search?query=email:bob.norman@mail.example.com) and [`GiftCard`](https://shopify.dev/docs/api/admin-rest/2024-01/resources/gift-card#get-gift-cards-search?query=last-characters:mnop) REST resources provide a `search` method, which are similar to the `all` method, in that they support pagination by cursor.

The information needed to make subsequent requests to a resource is returned in the response, but we're currently eliminating that information when returning information from those methods.

Added a new parameter to these methods, `returnObject`, which will return the full response when set, like `all`.

Open questions:
- Does `returnObject` work as the param name?
- I chose to make this a deprecating change in v10 since it's a much preferable way of returning this data, but it doesn't have to be - should we just make the param permanent?

## Type of change

## Checklist

### DIFF
--- a/packages/shopify-api/docs/guides/rest-resources.md
+++ b/packages/shopify-api/docs/guides/rest-resources.md
@@ -10,18 +10,19 @@ To make it easier to interact with the API, this library provides resource class
 
 ## Resource Methods
 
-| Resource Method | Description | Admin API endpoint | Return |
-| --------------- | ----------- | ------------------ | ------ |
-| `find` | Fetch a single resource by its ID | `GET /admin/api/{version}/{resource_name}/{resource_id}.json` | A single resource |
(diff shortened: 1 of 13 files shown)


paulomarg:
Requested a re-review just for the note in the docs!

PR #1266, opened 2024-03-07
Improve issue template

paulomarg:

We want to be able to solve as many problems as possible, and help unblock developers. However, when we get duplicate / hard to reproduce issues, it takes away from the time we could be spending fixing them.

Revamping the issue template to suggest some things developers can check before opening an issue to see if a solution has been found, or to ensure we're getting as much context as we can, which really makes it easier for us to solve issues.


nelsonwittwer (approved):
Great improvement!

PR #1288, opened 2024-03-19
Update the Session to return all user information

lizkenyon:

* First PR to start allowing saving of the Session user information. ([find current user](https://github.com/Shopify/shopify-app-template-remix/issues/366))

* Saving the user information will allow developers to introduce roll based access control for pages of their app, or log user activity.

* Update the toPropertyArray, and fromProperty array to handle the additional user information that was already on the Session object we got during the OAuth
* Once all of the Session Adapters have been updated we will deprecate the prior way of handling data that removes the additional user information. Currently both the new and old formats should both work in tandem.

## Type of change

## Checklist

### DIFF
--- a/.changeset/warm-hotels-allow.md
+++ b/.changeset/warm-hotels-allow.md
@@ -1,0 +1,59 @@
+---
+"@shopify/shopify-api": minor
+---
+
+Update session fromPropertyArray to handle all user info fields
+
+```tsconst {session, headers} = shopify.auth.callback({
+  rawRequest: req,
+  rawResponse: res,
+});
+/*
+  If session has the following data content...
+  {
+    id: 'online_session_id',
+    shop: 'online-session-shop',
+    state: 'online-session-state',
+    isOnline: true,
+    scope: 'online-session-scope',
+    accessToken: 'online-session-token',
+    expires: 2022-01-01T05:00:00.000Z,  // Date object
+    onlineAccessInfo: {
+      expires_in: 1,
+      associated_user_scope: 'online-session-user-scope',
+      associated_user: {
+        id: 1,
+        first_name: 'online-session-first-name',
+        last_name: 'online-session-last-name',
+        email: 'online-session-email',
+        locale: 'online-session-locale',
+        email_verified: true,
+        account_owner: true,
+        collaborator: false,
+      },
+    }
+  }
+ */
+
+const sessionProperties = session.toPropertyArray();
+/*
+  ... then sessionProperties will have the following data...
+   [
+    ['id', 'online_session_id'],
+    ['shop', 'online-session-shop'],
+    ['state', 'online-session-state'],
+    ['isOnline', true],
+    ['scope', 'online-session-scope'],
+    ['accessToken', 'online-session-token'],
+    ['expires', 1641013200000],  // example = January 1, 2022, as number of milliseconds since Jan 1, 1970
+    ['userId', 1],
+    ['first_name', 'online-session-first-name'],
+    ['last_name', 'online-session-last-name'],
+    ['email', 'online-session-email'],
+    ['locale', 'online-session-locale'],
+    ['email_verified', false]
+    ['account_owner', true,]
+    ['collaborator', false],
+    ],
+ */
+```
--- a/.changeset/warm-hotels-allow.md
+++ b/.changeset/warm-hotels-allow.md
@@ -1,13 +1,74 @@
 ---
 "@shopify/shopify-api": minor
 ---
+Updates the Session class to handle the associated user information on the session object.
 
-Update session fromPropertyArray to handle all user info fields
+Updates the Session `fromPropertyArray` to handle all user info fields.
 
-```tsconst {session, headers} = shopify.auth.callback({
+<details>
+
+```ts
+const sessionProperties = session.toPropertyArray(true);
+/*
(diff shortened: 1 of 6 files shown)


packages/shopify-api/lib/auth/oauth/types.ts
```diff
@@ -34,40 +34,42 @@ export interface OnlineAccessInfo {
...
+
+export interface OnlineAccessUser {
+  /**
+   * The user's ID.
+   */
+  id: number;
+  /**
+   * The user's first name.
+   */
+  first_name?: string;
```
lizkenyon, line 48: We currently have been using this OAuth type everywhere. 

From OAuth it will always have the full user information. 
When retrieving from storage, if they have not updated their session adapter then the additional user information will not be there.

To fix type errors with the tests I had to make these fields optional. Or should I make a new type for this?

packages/shopify-api/lib/session/__tests__/session.test.ts
```diff
@@ -271,13 +278,6 @@ const testSessions = [
         associated_user_scope: 'online-session-user-scope',
         associated_user: {
           id: 1,
-          first_name: 'online-session-first-name',
-          last_name: 'online-session-last-name',
-          email: 'online-session-email',
-          locale: 'online-session-locale',
-          email_verified: true,
-          account_owner: true,
-          collaborator: false,
```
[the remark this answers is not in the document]
lizkenyon: Removing these fields actually represents a new test case, where the session doesn't have the user information. Like the old way, after retrieving the data from the DB. 

I think all the the tests cases are still covered, but I am a little unclear on what each test scenario is

packages/shopify-api/lib/session/session.ts
```diff
@@ -205,14 +270,31 @@ export class Session {
...
           switch (key) {
             case 'expires':
-              return [key, value ? value.getTime() : undefined];
+              return [[key, value ? value.getTime() : undefined]];
             case 'onlineAccessInfo':
-              return [key, value?.associated_user?.id];
+              if (
+                value?.associated_user &&
+                Object.keys(value.associated_user).length === 1 &&
+                value.associated_user.id !== undefined
```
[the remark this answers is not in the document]
lizkenyon: 👍

packages/shopify-api/lib/auth/oauth/types.ts
lizkenyon: I added the new type in the latest commit. See if you are good with it!

packages/shopify-api/lib/session/types.ts
```diff
@@ -33,7 +33,22 @@ export interface SessionParams {
   /**
    * Information on the user for the session. Only present for online sessions.
    */
-  onlineAccessInfo?: OnlineAccessInfo;
+  onlineAccessInfo?: OnlineAccessInfo | StoredOnlineAccessInfo;
+}
+
+interface StoredOnlineAccessInfo {
```
paulomarg, line 39: Should this `extend OnlineAccessInfo`?

lizkenyon: Is this what you were thinking?
https://github.com/Shopify/shopify-api-js/pull/1288/commits/f8be3ada25db2cd96fbbb6d968a86b6bc89f92a6

.changeset/warm-hotels-allow.md
```diff
@@ -0,0 +1,114 @@
...
+        email_verified: false,
+        account_owner: true,
+        collaborator: false,
+      },
+    }
+  }
+ */
+```
+
+Updates the Session `toPropertyArray` to handle all user info fields. New optional argument `returnUserData`, (defaulted to `false`), will return the user data as part of property array object.
```
paulomarg, line 58: This seems to be the same as above?

.changeset/warm-hotels-allow.md
```diff
@@ -0,0 +1,114 @@
+---
+"@shopify/shopify-api": minor
+---
+Updates the Session class to handle the associated user information on the session object.
+
+Updates the Session `fromPropertyArray` to handle all user info fields. New optional argument `returnUserData`, (defaulted to`false`), will return the user data as part of the session object.
+
+```ts
```
paulomarg, line 8: I wonder if we want to put these examples inside a `<details>` element so it doesn't make the changelog too long. Although that makes me worry whether it will be visible enough.

packages/shopify-api/docs/guides/session-storage.md
```diff
@@ -65,6 +65,8 @@ Now that the app has a JavaScript object containing the data of a `Session`, it
 
 The `Session` class also includes an instance method called `.toPropertyArray` that returns an array of key-value pairs, e.g.,
 
+`toPropertyArray` has an optional parameter `returnUserData`, defaulted to false, when set to true it will return the associated user data as part of the property array object.
```
paulomarg, line 68: We should remember to update this to say this will be the way in the future when we do the deprecation.

paulomarg (approved):
Just the one point in the changeset.

.changeset/warm-hotels-allow.md
lizkenyon: I think I will put it in a details element.

paulomarg (approved):
🔥

PR #1298, opened 2024-03-26
Return user session data behind flag

lizkenyon:
Related to #1288 

Some of the adapters can pass in additional information so we need to add a flag to the `fromPropertyArray` to keep the same functionality, and remove this as possible breaking change.

* Adds an optional boolean flag to return the user session data

## Type of change

## Checklist


packages/shopify-api/lib/session/session.ts
```diff
@@ -85,55 +87,73 @@ export class Session {
                 },
               ];
             case 'userId':
-              return [
-                key,
-                (associatedUserObj.associated_user.id = Number(value)),
-              ];
+              if (returnUserData) {
```
lizkenyon, line 90: This switch statement got a more complicated with this change. Let me know if you are okay with this, or you would like this refactored into a (possibly) more readable if/else

paulomarg: I think that's probably fine - this will eventually be removed so it's not the end of the world.

paulomarg (approved):

PR #1318, opened 2024-04-02
feat: add Shop.current method

### LINKED ISSUE #1300: Rest resource `Shop.all()` returns the complete response
The call to `admin.rest.resources.Shop.all({ session })` returns the whole response, and not just the `data[0]` object:

https://github.com/Shopify/shopify-api-js/blob/eae4573cbe6991c4bda8044c1668dde9980bc9a3/packages/shopify-api/rest/admin/2024-01/shop.ts#L44

Like the `Customer` resurce, it should returns the following:

```typescript
return response.data ? response.data[0] : null;
```

nullndr:
- close #1300

## Changelog

This PR adds the new method `Shop.current()` that behaves the same as `Shop.all()` with the exception of unwrapping the response to return only the `data[0]` object like the method `Customer.find()`.

### DIFF
--- a/packages/shopify-api/rest/admin/2024-01/shop.ts
+++ b/packages/shopify-api/rest/admin/2024-01/shop.ts
@@ -28,6 +28,23 @@ export class Shop extends Base {
     }
   ];
 
+  public static async current(
+    {
+      session,
+      fields = null,
+      ...otherArgs
+    }: AllArgs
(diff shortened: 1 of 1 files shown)


paulomarg (approved):
Looks great! If you can please add a changeset (`yarn changeset` at the project root), I'll merge this :)

